PR #13, opened 2017-07-20
Integrate MondrianClient with Canon

jazzido:
Here's a sketch of a `<MondrianClientProvider>` component and its integration with the Canon example app.

Thanks!

### DIFF
--- a/app/components/App.jsx
+++ b/app/components/App.jsx
@@ -8,6 +8,10 @@ import {isAuthenticated} from "../../src";
 
 class App extends Component {
 
+  static contextTypes = {
+    mondrianClient: React.PropTypes.object.isRequired,
+  }
+
   constructor(props) {
     super(props);
   }
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "webpack": "^2.2.1",
     "webpack-dev-middleware": "^1.10.1",
     "webpack-hot-middleware": "^2.17.1",
-    "yn": "^2.0.0"
+    "yn": "^2.0.0",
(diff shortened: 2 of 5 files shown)


jazzido:
I'd like to be able to specify providers (i.e. top level elements that wrap the entire `App`)

From the app, somehow we need to pass an array like this to both`client.jsx` and `server.jsx` :

```javascript
const provs = [[ProviderOne, { prop: 1 }], [ProviderTwo, { otherProp: 2 }]]
```

Given this (which is present in `client` and `server.jsx`:

```jsx
const inner = <I18nextProvider i18n={i18n}>
    <Provider store={store}>
      <Router history={history} render={applyRouterMiddleware(renderMiddleware())}>
        {routes}
      </Router>
    </Provider>
  </I18nextProvider>
```
The following code would nest `inner` inside a nested structure of the passed providers.

```javascript
let c = provs.pop();
let comp = React.createElement(c[0], c[1], inner);
while ((c = provs.pop()) !== undefined) {
  comp = React.createElement(c[0], c[1], comp);
}
```

I'm not sure that the above code works (I didn't test it), but the idea is to programatically build something like this:

```jsx
<ProviderOne prop={1}>
<ProviderTwo otherProp={2}>
<I18nextProvider i18n={i18n}>
    <Provider store={store}>
      <Router history={history} render={applyRouterMiddleware(renderMiddleware())}>
        {routes}
      </Router>
    </Provider>
  </I18nextProvider>
</ProviderTwo>
</ProviderOne>
```

Issue #40, opened 2017-09-19
allow setting custom global constants


jazzido:
Feature request from the Datachile team (@palamago and yours truly): it would be awesome to be able to define constants at **build** time (e.g. with Webpack's `DefinePlugin`).

Our use case is instantiating `MondrianClient` with a parameter that depends on whether the app is running on the server or the client. Thus, we need to define both. Yes, environment variables, but they're not present as global constants; they only exist in the Redux store.

Also, to avoid writing yucky things like `if (window)…`, we'd love to have a `IS_SERVER` constant (trivially easy to do from Webpack's configuration).

davelandry:
Well isn't it your lucky day, I'm already using that plug-in (albeit very sparingly): https://github.com/Datawheel/datawheel-canon/blob/master/webpack/webpack.config.prod.js#L122

I think those variables should be refactored into two different Boolean variables: `__SERVER__` and `__DEV__`. Anything else you need in there for right now?

And also, to continue the idea, is there a preferred way of passing constants to there? Maybe through some environment variable prefix that I can scrape? Something like `CANON_CONST_*` gets transformed to `__*__`?

PR #55, opened 2017-12-05
Make environment vars CANON_CONST_* available as globals

jazzido:
Hey @davelandry,

Here's a first try at implementing what you suggested in #40.

Our (datachile) use case is creating a separate virtualhost for hosting static assets. That hostname would be defined as the envvar `CANON_CONST_ASSETS_HOSTS` .


davelandry:
@jazzido looks good to me!

PR #67, opened 2018-01-05
Fixes #66 — Resources need to be loaded at init() time.

### LINKED ISSUE #66: Trying to fix SSR and checksum mismatch
Hey,

Besides not taking advantage of SSR (except for SEO), the flickering caused by the mismatch between the server and client generated HTML is also a performance hit: you can _really_ see it in mobile.

Unfortunately, I don't have the time to go in and fix the issue, but here are some notes and hints that might help you solve it.

This is from the home page in the Canon example app:

image

The client version has a `<!-- react-empty: 15 -->` element right after the element with `data-reactid="14"` whereas the server version doesn't. However, when the page has finished rendering, the   `<!-- react-empty: 15 -->` comment _is not_ present in the browser's inspector. 

I don't think that we need React 16 to solve this. I downloaded, built and tested [the first two examples shown here](https://react.i18next.com/misc/universal-rendering.html), and they work perfectly with React 15 (no "checksum" message). So it's likely that our usage of i18next is causing this. I removed all usages of i18next components from `Home.jsx`, and I didn't get any errors:

```jsx
import React, {Component} from "react";
import {Link} from "react-router";
import {translate, Trans} from "react-i18next";
import "./Home.css";

import {Activate} from "../../src/";

// import styles from "style.yml";

// import yn from "yn";

class Home extends Component {

  render() {
    const {t} = this.props;
    const count = 5, name = "Dave";
    // console.log("yn test", yn("true"));
    // console.log(styles);
    // use the 't' function for simple keys and strings, and
    // the Interpolate component for strings with variable replace.
    return (
      <div className="home">
        This is some static text
      </div>
    );

  }
}
export default Home;
```

This is not a blocker, but performance would be a lot better if we fix it.

Thanks!

jazzido:
I'm 98.74% sure that this fixes the checksum mismatch issue. Works with canon app, also with Datachile (`npm link`ed canon into datachile)

I'd appreciate a review, test, merge and release.

Thanks.

### DIFF
--- a/src/client.jsx
+++ b/src/client.jsx
@@ -31,10 +31,11 @@ const {CANON_LOGLOCALE, NODE_ENV} = window.__INITIAL_STATE__.env;
 // do rewrite this monstrosity. I hate JS anyway.
 const res = {};
 res[locale] = {};
-res[locale][window.__APP_NAME__] = {
-  ...defaultTranslations,
-  ...resources
-};
+res[locale][window.__APP_NAME__] = resources;
+
+res['canon'] = {};
+res['canon'][window.__APP_NAME__] = defaultTranslations;
+
 
 i18n
   .init({
--- a/src/client.jsx
+++ b/src/client.jsx
@@ -26,6 +26,16 @@ import Canon from "./Canon";
 const {locale, resources} = window.__INITIAL_STATE__.i18n;
 const {CANON_LOGLOCALE, NODE_ENV} = window.__INITIAL_STATE__.env;
 
+
+// XXX If there's a better way to build an object with variable keys, please
+// do rewrite this monstrosity. I hate JS anyway.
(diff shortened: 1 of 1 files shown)


jspeis:
I saw the comments in the commit and FWIW you could use ES6's bracket syntax to create an object with variable keys all at once
e.g.

screen shot 2018-01-05 at 7 51 13 am

davelandry:
@jazzido wow! great sleuthing here. I'll be merging this now and using @jspeis' ES6 suggestion for object construction

davelandry:
for posterity, here's the new object format:

```js
const name = window.__APP_NAME__;

const res = {
  [locale]: {[name]: resources},
  canon: {[name]: defaultTranslations}
};
```

PR #84, opened 2018-04-03
Google OAuth integration

jspeis:
This PR adds the core elements needed for CNY to integrate with Google's login infrastructure:

* Adds the Google passport module (passport-google-auth20)
* Updates DB table structure to reference Google
* Adds placeholder localization strings
* Adds basic Google button and css (see note below)
* Adds Google logo
* Basic setup instructions

One other thing this PR does is that it also explicitly prefixes social OAuth callback URLs with `CANON_SOCIAL_REDIRECT` URL. Most providers seem to (now?) be strict about this (_new_ Apps on Google and Facebook, for example, don't seem to work without the fully qualified callback URL now). 

@davelandry I threw in some placeholder CSS but would be great if you could give that a real pass 😄


davelandry:
@jspeis aaaaand the button looks great. I'm not touching it.

screen shot 2018-04-03 at 10 27 47 am

jspeis:
Awesome thanks! 😀

PR #88, opened 2018-04-09
Hotfixes social buttons import path for external projects

jspeis:
It looks like I caused a bug in the social buttons components due to the module import path. It would work when testing within canon but when trying to use the components outside of canon from another project the import path gave errors. These changes appear to fix that.


jspeis:
updated!

PR #122, opened 2018-07-12
Vizbuilder 0.1

### LINKED ISSUE #105: switch measure select to use virtualized list
I've used this library in the past: https://github.com/bvaughn/react-virtualized-select

### LINKED ISSUE #108: split "level" selector into 2 new selectors
A better user behavior would be as follows:

### First dropdown:
* title: "grouped by"
* values: dimension groups (for example, "Geography" or "PUMS Industry")
* component: single-selector

### Second dropdown:
* title: "at depth level"
* values: the available dimension members based on the previous dropdown selection (for example, "State" or "County" if Geography has been selected)
* component: single-selector

This also means that we are removing the ability (for now) to choose multiple dimensions. I would still make the Year dimension always applied, but we do not need to show it in the interface.

### LINKED ISSUE #118: group measures by cube topic and subtopic
we need to find a measure selector that allows for multiple grouping levels, with the idea being instead of 1 big flat list we could do this:

- "Health"
  - "Access and Quality"
    - "Adult Patients Who Did Not Receive Good Communication About Hospital Discharge"
    - ...

@cnavarreteliz @frabarz I haven't done any real deep searching for components (probably related to #105)... have you guys seen anything that could allow that?

frabarz:
First stable version for the vizbuilder.
Internal components were designed for extendability, so future changes should not extend too much through the codebase.

Closes #105 
Closes #108 
Closes #118


packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+  );
+
+  const measureName = query.measure.name;
+  const getMeasureName = d => d[measureName];
+
+  const commonConfig = {
+    height: activeType ? 500 : 400,
+    legend: false,
+
+    tooltip: true,
```
davelandry: remove this, tooltips are always on by default.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+  keyTitle = keyTitle || query.drilldown.name;
+  keyBody = keyBody || query.measure.name;
+
+  const config = {
+    width: 60,
+    title: d =>
+      `<h5 class="title xs-small">${[].concat(d[keyTitle]).join("<br/>")}</h5>`
+  };
+
+  if (query.timeDrilldown && keyTitle !== "Year") {
```
davelandry: rewrite this big `if` statement to use the tooltip's built-in table, and conditionally add more arrays to the list if the values (like moe and year) are present. here is an example usage: https://d3plus.org/examples/d3plus-tooltip/getting-started/

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+
+  const colorScale = {
+    colorScale: measureName,
+    colorScalePosition: "bottom",
+    colorScaleConfig: {width: 0, height: 0}
+  };
+
+  if (!activeType) {
+    const hasTimeDim = availableKeys.has("Year");
+    const hasGeoDim = availableKeys.has("State") || availableKeys.has("County");
```
davelandry: we can't rely on adding catches for each geography type for different cubes. I have proposed to @jspeis and @hwchen adding a `"GEOGRAPHY"` annotation for us to detect this: https://github.com/Datawheel/datausa-tracker/issues/77

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+    }
+    if (aggregatorType !== "SUM") {
+      availableCharts.delete("treemap");
+      availableCharts.delete("barchartyear");
+    }
+  }
+  else {
+    if (year !== ALL_YEARS) {
+      commonConfig.total = getMeasureName;
+      commonConfig.totalConfig = {
```
davelandry: remove this totalConfig, as it is styling that should be the user's responsibility.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+
+    return config;
+  },
+  geomap(commonConfig, query, flags) {
+    const drilldownName = query.drilldown.name;
+
+    const config = {
+      ...commonConfig,
+      ...flags.colorScale,
+      tiles: false,
```
davelandry: remove this `tiles: false`, let the user of the vizbuilder decide if they want tiles.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+      id: `ID ${drilldownName}`,
```
davelandry: remove this, there is no `id` method of d3plus.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+      groupBy: `ID ${drilldownName}`,
+      ocean: "transparent",
```
davelandry: remove `ocean: "transparent"`, let the user of the vizbuilder decide this.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+      projection: "geoAlbersUsa",
```
davelandry: remove `projection: "geoAlbersUsa"`, let the user of the vizbuilder decide this.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+      topojson: "/topojson/states.json",
+      topojsonId: "id",
+      topojsonKey: "states",
+      zoom: true,
+      zoomFactor: 2,
```
davelandry: remove `topojsonId`, `topojsonKey`, `zoom`, and `zoomFactor`, let the user of the vizbuilder decide this.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+
+    const config = {
+      ...commonConfig,
+      ...flags.colorScale,
+      tiles: false,
+      id: `ID ${drilldownName}`,
+      groupBy: `ID ${drilldownName}`,
+      ocean: "transparent",
+      projection: "geoAlbersUsa",
+      topojson: "/topojson/states.json",
```
davelandry: we need to generalize this. how about this solution:

The user of the `Vizbuilder` component can pass a lookup object to props that is a mapping of dimension names to topojson files. We will only draw Geomaps if we have a correct topojson mapping for the dimension selected. For example:

```jsx
<Vizbuilder src="https://canon-api.datausa.io"
  topojson={{
    State: "/topojson/states.json",
    County: "/topojson/counties.json"
  }} />
```

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+    totalConfig: {
+      fontSize: 14
+    }
+  };
+
+  const colorScale = {
+    colorScale: measureName,
+    colorScalePosition: "bottom",
+    colorScaleConfig: {width: 0, height: 0}
+  };
```
davelandry: I think we should remove this extra config object, don't pass it through `flags`, and just allow each visualization config to set the `colorScale: measureName`.

`colorScalePosition` should be set by the user, and `colorScaleConfig` should not have to be set anymore (I fixed `colorScalePosition: false`)

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+    };
+
+    if (moe) {
+      const moeName = moe.name;
+
+      config.confidence = [
+        d => d[measureName] - d[moeName],
+        d => d[measureName] + d[moeName]
+      ];
+      config.confidenceConfig = {
```
davelandry: this `confidenceConfig` should be user-definable, please remove.

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+    const {drilldown, measure} = query;
+
+    const drilldownName = drilldown.name;
+    const measureName = measure.name;
+
+    const config = {
+      ...commonConfig,
+      groupBy: drilldownName,
+      y: measureName,
+      yConfig: {title: measureName}
```
davelandry: `y` and `yConfig` are not needed here.

packages/vizbuilder/src/components/AreaChart.jsx
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  dispatchScroll() {
+    // TODO: Discuss how could we implement IntersectionObserver
+    window.dispatchEvent(new CustomEvent("scroll"));
+  }
+
+  dispatchResize() {
+    window.dispatchEvent(new CustomEvent("resize"));
+  }
```
davelandry: what do these `"scroll"` and `"resize"` events do? I believe they are not necessary.

packages/vizbuilder/src/components/ChartCard.jsx
```diff
@@ -0,0 +1,47 @@
...
+      config,
+      dataset,
+      onSelect,
+      query,
+      type,
+      yearSelector
+    } = this.props;
+    const ChartComponent = charts[type];
+
+    const title = `${type} of ${query.measure.name || ""} by ${query.drilldown
```
davelandry: move this title logic into the chartconfig logic

packages/vizbuilder/.editorconfig
```diff
@@ -0,0 +1,15 @@
+# http://editorconfig.org
```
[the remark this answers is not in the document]
frabarz: The .editorconfig file is an attempt to standardize all these config files for every text editor/IDE on the market. It doesn't allow for many options though. https://editorconfig.org/

packages/vizbuilder/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@datawheel/vizbuilder",
+  "description": "Visualization explorer for data retrieved from mondrian-rest.",
+  "version": "0.1.0",
+  "main": "src/index.js",
+  "author": "Francisco Abarzúa & Carlos Navarrete",
+  "license": "MIT",
+  "typings": "index.d.ts",
```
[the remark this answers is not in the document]
frabarz: The *.d.ts files are typescript files, but their function is to provide typing information about functions for javascript modules and libraries. Since javascript doesn't work with typing like typescript does, module owners can create a *.d.ts file and add it to their package.json, so VisualStudio Code and other editors are able to parse and provide static typing verification.

Some editors are able to parse JSDoc and do the same (VSCode actually can), but this functionality is limited when you try to import a module from a npm package. This file can surpass that limitation.

packages/vizbuilder/index.d.ts
```diff
@@ -0,0 +1,18 @@
+import {Cube} from "mondrian-rest-client";
+import Dimension, {Level} from "mondrian-rest-client/lib/types/dimension";
+import Measure from "mondrian-rest-client/lib/types/measure";
+import { IVizbuilderState, IVizbuilderProps } from "./typings/state";
```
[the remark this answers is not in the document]
frabarz: Yeah, I only made some attempts on working with this file, but in the end I couldn't configure it correctly. I actually wasn't intending to upload it to the repo, it just slipped. I'll delete the references.

packages/vizbuilder/src/index.js
```diff
@@ -0,0 +1,112 @@
+import React from "react";
+import PropTypes from "prop-types";
+import classnames from "classnames";
+import queryString from "query-string";
+
+import {fetchCubes} from "./actions/fetch";
+import {loadControl, setStatePromise} from "./actions/loadstate";
+import AreaChart from "./components/AreaChart";
+import AreaLoading from "./components/AreaLoading";
```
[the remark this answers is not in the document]
frabarz: Good catch, is basically the same code. However canon's Loading includes the translate HOC and the redux plug, and so it uses redux's values for the loading progress.
The vizbuilder is pretty much a capsuled app, all the data, including the progress counter and loading state, is inside the main component. In its current state, the vizbuilder could run both in a canon app or in a create-react-app app. Also, adding canon as a dependency of vizbuilder only to use its Loading component is a bit overkill.

packages/vizbuilder/src/state.js
```diff
@@ -0,0 +1,77 @@
...
+  orderDesc: PropTypes.bool,
+  timeDrilldown: PropTypes.any
+});
+
+export const optionsTypes = PropTypes.shape({
+  cubes: PropTypes.array,
+  dimensions: PropTypes.array,
+  levels: PropTypes.array,
+  measures: PropTypes.array
+});
```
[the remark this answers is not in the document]
frabarz: These are a residue of when I tried to pass the values through context, and were also part of an experiment on static typing enforcement that didn't work. I'll remove them.

packages/vizbuilder/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+Vizbuilder.childContextTypes = {
+  loadControl: PropTypes.func,
+  stateUpdate: PropTypes.func
+};
+
+Vizbuilder.propTypes = {
+  src: PropTypes.string,
+  datasetWillChange: PropTypes.func,
+  datasetDidChange: PropTypes.func
```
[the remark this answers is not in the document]
frabarz: Those are lifecycle functions I'm planning to add.
When working on datachile, we sometimes had to hack our way into some modules to get certain information. Having lifecycle hooks available is always a good idea, even more in a capsuled component like vizbuilder, since there's no other way to get the internal state.
For example, if I need to access the current dataset from the parent app to keep a history, or to hook the current state and update the permalink.

packages/vizbuilder/src/helpers/query.js
```diff
@@ -0,0 +1,123 @@
...
+//     // when parents requested, also get their i18n'd captions
+//     if (query.options.parents) {
+//       rangeRight(1, level.depth).forEach(d => {
+//         const ancestor = level.hierarchy.levels.find(l => l.depth === d);
+//         query = setCaptionForLevelAndLang(query, ancestor, lang);
+//       });
+//     }
+//   });
+//   return query;
+// }
```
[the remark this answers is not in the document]
frabarz: will do

packages/vizbuilder/src/components/ChartCard.jsx
frabarz: this title is currently not rendered in the chart, but in the component, along with the year selector
should I move this title to the chart itself? otherwise it doesn't make sense to move this logic

packages/vizbuilder/src/components/AreaChart.jsx
frabarz: Since the vizbuilder uses a lot of charts at the same time, and there's a transition to change the size of the extended chart, sometimes the other charts aren't rendered after the user comes back to the multi-chart view. Also, when there's only 2 charts in a view, there's no scrollbar to ensure rendering if that happens, so these 2 functions are there as safeguards.
They're used only in specific `div`s, and have debouncers, so the overhead is minimal.

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: Corrected in fdaed77616abce7ffcc4458f69e029428c96d9af

packages/vizbuilder/src/state.js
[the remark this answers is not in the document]
frabarz: Removed in ce3b1ae369befac12d261191ad069dbe5bf8c39a

packages/vizbuilder/index.d.ts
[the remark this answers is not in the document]
frabarz: Removed in ce3b1ae369befac12d261191ad069dbe5bf8c39a

packages/vizbuilder/package.json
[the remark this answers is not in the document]
frabarz: Removed in ce3b1ae369befac12d261191ad069dbe5bf8c39a, though I think we should add it at some point in the future.

packages/vizbuilder/.editorconfig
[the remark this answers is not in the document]
davelandry: ahh interesting, this file is ok to stay in the repository (until we maybe add a standard version to canon)

packages/vizbuilder/src/components/ChartCard.jsx
davelandry: yes, moving the titles into the chart itself will also help with accessibility, as this will become a hidden `<title>` element inside of the SVG

packages/vizbuilder/src/helpers/random.js
```diff
@@ -0,0 +1,29 @@
...
+export function pickOne(list) {
+  return list[Math.floor(Math.random() * list.length)];
+}
+
+/**
+ * Slices a string to
+ * @param {string} string The string to slice
+ * @returns {string}
+ */
+export function sortSlice(string) {
```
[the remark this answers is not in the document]
frabarz: This function is actually very important.
The virtualized list in the selector doesn't actually sort the elements it's showing, the elements are sorted just once, just before they're saved in the vizbuilder's state. The virtualized list only makes the hierarchy iterating over the title properties, but if they were unsorted, we would have repeated titles.
This function makes a kind of index key for the sort function, and it needs the slicing because of ASCII.
I ended up removing the other two functions, so I'll move this one next to the function that uses it, and delete random.js.

packages/vizbuilder/src/helpers/sorting.js
```diff
@@ -0,0 +1,211 @@
...
+  while (nCbs--) {
+    const cube = cubes[nCbs];
+    let nMsr = cube.measures.length;
+    while (nMsr--) {
+      const measure = cube.measures[nMsr];
+      // TODO: ensure this is checked server-side and remove
+      if (/\smoe$/i.test(measure.name)) {
+        continue;
+      }
+      const key = measure.annotations.error_for_measure;
```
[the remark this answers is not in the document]
frabarz: That's the reason for the `// TODO`
For test purposes, and since this annotation is not completely applied to the main cube server, this works as a safeguard.

packages/vizbuilder/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+import PropTypes from "prop-types";
+import classnames from "classnames";
+import queryString from "query-string";
+
+import {fetchCubes} from "./actions/fetch";
+import {loadControl, setStatePromise} from "./actions/loadstate";
+import AreaChart from "./components/AreaChart";
+import AreaLoading from "./components/AreaLoading";
+import AreaSidebar from "./components/AreaSidebar";
+import {UIToaster} from "./components/UIToaster";
```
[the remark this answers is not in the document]
frabarz: done in e91ee08ac4a0ad5edfec7314cac7069b5035893b

packages/vizbuilder/src/helpers/sorting.js
```diff
@@ -0,0 +1,211 @@
...
+      annotations._cb_subtopic = cbSubtopic;
+      annotations._cb_sourceName = sourceName;
+      annotations._selectorKey =
+        selectorKey + (measure.caption || measure.name);
+      annotations._sortKey = `${sortKey}${measure.caption ||
+        measure.name}`.toLowerCase();
+      // annotations._source_desc = sourceDesc;
+      // annotations._source_link = sourceLink;
+      // annotations._dataset_name = datasetName;
+      // annotations._dataset_link = datasetLink;
```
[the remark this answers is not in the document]
frabarz: done in 8b1478042fbee468dc59cb006b3a01e8812c211a

packages/vizbuilder/src/helpers/random.js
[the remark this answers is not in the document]
frabarz: done in dff1dcb395fac24455ec5272afcaa2bc3e2bea20

packages/vizbuilder/src/helpers/random.js
```diff
@@ -0,0 +1,29 @@
...
+    .substr(2, 5)}`;
+}
+
+/**
+ * Chooses one random element from an array.
+ * @template T
+ * @param {Array<T>} list Array with the elements from where to pick one
+ * @returns {T}
+ */
+export function pickOne(list) {
```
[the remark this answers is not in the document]
frabarz: done in dff1dcb395fac24455ec5272afcaa2bc3e2bea20

packages/vizbuilder/src/helpers/random.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Returns a random, non-cryptographically secure, unique id string.
+ * @returns {string}
+ */
+export function makeRandomId() {
```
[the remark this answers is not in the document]
frabarz: done in dff1dcb395fac24455ec5272afcaa2bc3e2bea20

packages/vizbuilder/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+    return (
+      <div className={classnames("vizbuilder", {loading: load.inProgress})}>
+        <AreaLoading {...load} />
+        <AreaSidebar options={options} query={query} queryOptions={queryOptions} />
+        <AreaChart dataset={dataset} members={members} query={query} />
+      </div>
+    );
+  }
+
+  stateUpdate(newState) {
```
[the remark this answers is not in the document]
frabarz: Nope, because this code is combining elements exactly one level deep in the state.
If `newState` is `{query: {conditions: newConditions}}`, this code merges the elements inside `query`.

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: ...alright? ~but... how... does...?~

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+      y: measureName,
+      yConfig: {title: measureName},
+      stacked: drilldown.depth > 1
+    };
+
+    if (timeDrilldown) {
+      config.groupBy = [timeDrilldown.hierarchy.levels[1].name];
+    }
+
+    if (flags.availableKeys.has("State")) {
```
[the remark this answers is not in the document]
frabarz: That part of the code set a sorting function for the elements on the x axis, should I keep the way is it set now (basically, `localeCompare` between drilldown names), or should I remove it?

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+      topojson: "/topojson/states.json",
+      topojsonId: "id",
+      topojsonKey: "states",
+      zoom: true,
+      zoomFactor: 2,
+      zoomScroll: false
+    };
+
+    if (flags.activeType === "geomap") {
+      const size = getAreaChartDimensions();
```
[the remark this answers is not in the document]
frabarz: you're right, in the end I removed it in 50f17d1844e8f061476b4c16dc722d7a0701b951

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+    legend: false,
+
+    tooltip: true,
+    tooltipConfig: tooltipConfig(query),
+
+    duration: 0,
+    loadingMessage: "Loading",
+    sum: getMeasureName,
+    value: getMeasureName,
+    totalConfig: {
```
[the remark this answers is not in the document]
frabarz: done in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,330 @@
...
+
+  const commonConfig = {
+    height: activeType ? 500 : 400,
+    legend: false,
+
+    tooltip: true,
+    tooltipConfig: tooltipConfig(query),
+
+    duration: 0,
+    loadingMessage: "Loading",
```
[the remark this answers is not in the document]
frabarz: done in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: done in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: done in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: removed in 714f01f39181d4f723035741609a115dcab42333

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: removed in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: removed in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: removed in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: Considering the need for chart config from the user, for now I set it as another attribute, `config`, which receives an object with 2 keys:
* `topojson`, whose purpose is provide this config
* `chartConfig`, which provides the config for the charts
The detail of the structure are available in the [`propTypes` property](https://github.com/Datawheel/datawheel-canon/blob/096b8f5a0900b84d4341cca21e35db479343399a/packages/vizbuilder/src/index.js#L121), and an example is available in the [`defaultProps` property](https://github.com/Datawheel/datawheel-canon/blob/096b8f5a0900b84d4341cca21e35db479343399a/packages/vizbuilder/src/index.js#L154), both in the main Vizbuilder component.

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: removed in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: removed in 096b8f5a0900b84d4341cca21e35db479343399a

packages/vizbuilder/src/index.js
```diff
@@ -0,0 +1,198 @@
...
+      stackedActive: PropTypes.object,
+      treemap: PropTypes.object,
+      treemapActive: PropTypes.object
+    })
+  }),
+  src: PropTypes.string.isRequired
+};
+
+// EXAMPLE, will be removed later
+Vizbuilder.defaultProps = {
```
davelandry: I tried a simple copy/paste of the vizbuilder into my Data USA test site, and am having problems overriding some of these configs. I suggest we make a couple of changes:
1. a property called `topojson`, which like you have it, accepts keys for cube `Level` names. The configuration for each level should simply be folded into the default config that is constructed for Geomaps: `{..geoMapConfig, ...topojsonConfig[level]}`. Default should be `{}`, which means no Geomaps out of the box because they rely on static topojson and we can't predict where the user could be storing it.
2. a property called `config`, which is simply a standard d3plus config object (but defaults to and empty `{}` object). Most configs will/should be set in `app/d3plus.js`, which gets applied to all `d3plus-react` components automatically through context, but I can see a scenario where the vizbuilder may need unique styles
3. a property called `visualizations`, which should be an array of string values matching the visualization names defined in your `charts` variable. this would allow the user to toggle the visibility of specific visualization types. By default, do not show `Pie` or `Donut` (`Treemap` is our preferred "share" visualization)

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,359 @@
+import {
+  BarChart,
+  Donut,
+  Geomap,
+  LinePlot,
+  StackedArea,
+  Treemap
+} from "d3plus-react";
```
davelandry: add support for `Pie`

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,334 @@
...
+export const ALL_YEARS = "All years";
+
+export const tooltipGenerator = (label, drilldowns, measure, moe) => {
+  const tbody = Array.from(drilldowns)
+    .filter(d => d !== label)
+    .map(dd => [dd, d => d[dd]]);
+  tbody.push([measure, d => `${(d[measure] * 1 || 0).toFixed(4)}`]);
+  if (moe) {
+    moe = moe.name;
+    tbody.push(["Margin of Error", d => `±${(d[moe] * 1 || 0).toFixed(4)}`]);
```
davelandry: For formatting these tooltip values, use `formatAbbreviate` from `d3plus-format`

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,334 @@
...
+  );
+
+  const drilldownName = query.drilldown.name;
+  const measureName = query.measure.name;
+  const getMeasureName = d => d[measureName];
+
+  const commonConfig = {
+    height: activeType ? 500 : 400,
+    legend: false,
+
```
davelandry: When viewing tooltips, I'm seeing the drilldown type and the value (ie. "State California"). It should just be the value. Maybe the `commonConfig` should have `label: dimensionName`?

packages/vizbuilder/src/helpers/chartconfig.js
```diff
@@ -0,0 +1,334 @@
...
+    return config;
+  },
+  geomap(commonConfig, query, flags) {
+    const drilldownName = query.drilldown.name;
+    const measureName = query.measure.name;
+    const isActive = flags.activeType === "geomap";
+
+    const topojsonConfig =
+      flags.topojson[drilldownName] || flags.topojson.default;
```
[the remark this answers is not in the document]
frabarz: updated in e14e6928e360e73ca21b8a5ff34a61f8af76a2f4

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: applied in 48bb01690be619a1e59f2e1c72a48707ddd723fb

packages/vizbuilder/src/helpers/chartconfig.js
frabarz: applied in f909c70bed94a0d7bd8c57d085cc500e6c140875

PR #140, opened 2018-08-21
[vizbuilder] tick formatting in yConfig

### LINKED ISSUE #133: [vizbuilder] apply tickFormat to yConfig
By default, this should use the `formatAbbreviate` function exported from `d3plus-format`, and if the measure's `"units_of_measurement"` is set to "Dollars", then add a dollar sign before the value

frabarz:
Closes #133


packages/vizbuilder/src/index.js
```diff
@@ -137,6 +143,9 @@ Vizbuilder.propTypes = {
 
 Vizbuilder.defaultProps = {
   config: {},
+  formatting: {
+    Dollars: d => `$${formatAbbreviate(d * 1 || 0)}`
+  },
```
davelandry: If the user supplies their own formatters for other annotations, will that delete the preset here? If so, we have to implement this in a way so that if a user passed `formatting={{Time: d => "this is time"}}` it wouldn't delete the default `Dollars` formatter.

frabarz:
applied on 0fc3822a8ecb962ac626a916a5ba6d9532ddf109

davelandry:
@frabarz can you take a look at the merge conflicts here? I'll rebase after they are resolved.

frabarz:
sorry, I didn't notice
the branch is now ready

PR #168, opened 2018-09-11
Uses log scale when necessary

### LINKED ISSUE #148: [vizbuilder] use log scales when necessary
In Line Plot, when sqrt(SD)/average be much higher than 1, use log scale by default.

cnavarreteliz:
Closes #148


packages/vizbuilder/src/helpers/math.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Calculate the average of a data array
+ * This means it should have a numeric value, and a valid operator.
+ * @param {Array} data An array to check
+ * @param {String} measureName Name of the measure
+ */
+export function average(data, measureName) {
```
davelandry: can we just use `mean` from `d3-array`? https://github.com/d3/d3-array#mean

packages/vizbuilder/src/helpers/math.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+
+/**
+ * Calculate the standard deviation of a data array
+ * This means it should have a numeric value, and a valid operator.
+ * @param {Array} data An array to check
+ * @param {String} measureName Name of the measure
+ */
+export function standardDeviation(data, measureName) {
```
davelandry: can we just use `deviation` from `d3-array`? https://github.com/d3/d3-array#deviation

packages/vizbuilder/src/components/ChartArea/ChartCard.jsx
```diff
@@ -16,6 +17,11 @@ class ChartCard extends React.Component {
 
     config.data = dataset;
 
+    if (type === "lineplot" && relativeStdDev(dataset, config.y) > 1) {
+      config.yConfig.scale = "log";
+      config.yConfig.gridConfig = {"stroke-width": 0};
```
davelandry: If the use defines a specific `gridConfig` (for maybe something other than "stroke-width") and `scale` using the `config` prop, do these overwrite those values? We should only manually set these settings if they have not been defined by the user (or apply the user's config after this)

cnavarreteliz: I think that we can fix it in d3plus. I'm going to create an issue.

PR #181, opened 2018-09-24
adds support for new annotations

cnavarreteliz:
closes #190


davelandry:
@cnavarreteliz does this also hide these measures from the measure selector?

cnavarreteliz:
yes @davelandry :) but doesn't add the `source` and `collection` annotations to the tooltips yet.

cnavarreteliz:
@frabarz I'm ok with my changes

PR #195, opened 2018-09-26
[vizbuilder] adds support metacharts in d3plus

cnavarreteliz:


packages/vizbuilder/src/components/ChartArea/index.jsx
```diff
@@ -103,7 +103,7 @@ class ChartArea extends React.Component {
         activeChart,
         generalConfig
       );
-      const asdf = chartConfigs.map((chartConfig, i) => (
+      const asdf = chartConfigs.map(chartConfig => (
```
davelandry: @frabarz you let this variable name through!? ;-)

PR #224, opened 2018-10-12
[vizbuilder] increases TooMuchData default cap

### LINKED ISSUE #223: [vizbuilder] increase TooMuchData cap and make env var
Currently, I'm unable to create a county map in the vizbuilder, because ~3221 US Counties times the number of years available is over the `9999` limit for too much data. Ideally, we shouldn't need this message and our visualizations should be able to handle everything we try to do, but for now let's increase this limit to something like `20000` and also allow the user to set this using an env var: `CANON_VIZBUILDER_DATACAP`

If there was any way to take the number and divide it by the number of years, that would also be a great solution to help with this.

frabarz:
closes #223 

I'm not really sure if this is how I should handle the environment variable though. Won't the client have issues with `process.env` or is it handled by webpack?

### DIFF
--- a/packages/vizbuilder/src/helpers/fetch.js
+++ b/packages/vizbuilder/src/helpers/fetch.js
@@ -10,8 +10,6 @@ import {
   getValidMeasures
 } from "./sorting";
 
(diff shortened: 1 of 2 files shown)


frabarz:
alright, now it handles it through component attribute
the code also makes the calculation dividing by amount of years

PR #252, opened 2018-10-22
implements sticky headers in measure select in vizbuilder

### LINKED ISSUE #247: [vizbuilder] sticky headers in custom selector
It's currently very easy to loose context while scrolling through the measure list (ie. "are these education measures? I forget"). The scrolling list should have sticky headers where the currently viewed grouping stays at the top of the scrolling area, similar to this: https://codepen.io/gnettles/pen/xrlqI/

frabarz:
this implementation uses `position: sticky` to set the trick
afaik, Safari has limited availability for that property, but the other browsers provide enough coverage

closes #247


davelandry:
@frabarz looks great!

PR #253, opened 2018-10-24
[vizbuilder] implements measure grouping by table_id annotation

### LINKED ISSUE #249: [vizbuilder] group measures by cube table_id
- [x] remove `dataset_name` from the measure selector subtitle
- [x] group measures together if their cubes' `table_id` annotation matches
- [x] if there is only one cube for a measure, keep the currently displayed `Dataset: {dataset_name}` text underneath the measure selector. if there are multiple cubes, create a dropdown where you select between the cubes, showing the `dataset_name` annotation as the name of the cube.

frabarz:
closes #249

### DIFF
--- a/packages/vizbuilder/src/components/Sidebar/DatasetSelect.jsx
+++ b/packages/vizbuilder/src/components/Sidebar/DatasetSelect.jsx
@@ -1,0 +1,28 @@
+import React from "react";
+
+function DatasetSelect(props) {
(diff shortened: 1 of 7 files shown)


davelandry:
looks great @frabarz! I made one small commit that adds a "pt-select" div wrapper around the DatasetSelect component so we get some nice default blueprint styles

PR #268, opened 2018-11-03
[vizbuilder] implements sidebar related features

### LINKED ISSUE #264: [vizbuilder] add top/bottom list to sidebar
Similar to the [current DataUSA map page](https://datausa.io/map/?level=county&key=uninsured), we should display a list of top/bottom values from the data underneath the filters but above the source information.

If the dataset has less than 20 values, display as one list labeled "Rankings". If there are more than 20 values, display "Top 10" and "Bottom 10" lists.

In regards to what name to display, as there can be multiple groupings, let's try this format: join the group values using `" - "`.

For example, with [this](https://ironwood.datausa.io/visualize?groups=0-z9TnC&groups=1-ZgMrcP&measure=1qWfo) query, the title would look like: `"Massachusetts - High School Equivalent"

### LINKED ISSUE #265: [vizbuilder] allow React component to accept children
For example, the user may want to add additional controls/filters (I wanted to add an "Add to Data Cart" button) to the sidebar. If the user uses passes children to the component, render `this.props.children` directly underneath the "Add Filter" button (but before the source info and top/bottom lists mentioned [here](https://github.com/Datawheel/canon/issues/264))

Example usage:

```jsx
<Vizbuilder src="url">
  <button onClick={myFunction}>Add to Data Cart</button>
</Vizbuilder>
```

### LINKED ISSUE #266: [vizbuilder] text wrap data source in measure selector
The measure names wrap nicely, but the sources do not:

screen shot 2018-11-02 at 12 18 18 pm

### LINKED ISSUE #267: [vizbuilder] activeChart never appearing in onChange payload
When hooking into the `onChange` event, there is a field called `activeChart` that is always undefined. My expectation would be that the `onChange` event gets fired when the user chooses to enlarge/close a visualization, and that field would change to show the enlarged visualization.

frabarz:
closes #264
closes #265 
closes #266 
closes #267

### DIFF
--- a/packages/vizbuilder/src/components/ChartArea/index.jsx
+++ b/packages/vizbuilder/src/components/ChartArea/index.jsx
@@ -8,18 +8,6 @@ import ChartCard from "./ChartCard";
 
 import "./style.css";
 
-const DEFAULT_FORMATTERS = {
-  Growth: d => `${formatAbbreviate(d * 100 || 0)}%`,
-  Percentage: d => `${formatAbbreviate(d * 1 || 0)}%`,
-  Rate: d => `${formatAbbreviate(d * 100 || 0)}%`,
-  Ratio: d => `${formatAbbreviate(d * 1 || 0)} to 1`,
-  USD: d => `$${formatAbbreviate(d * 1 || 0)}`,
-  get Dollars() { return this.USD },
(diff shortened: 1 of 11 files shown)


davelandry:
Great work on these improvements @frabarz!

PR #326, opened 2018-11-30
changes render behavior for perc bar for a cdc chart

jhmullen:
This adds a `showPercent` switch to the data config for perc bars so I can properly render one of the charts for CDC

### DIFF
--- a/packages/cms/src/components/Viz/PercentageBar.jsx
+++ b/packages/cms/src/components/Viz/PercentageBar.jsx
@@ -16,7 +16,10 @@ class PercentageBar extends Component {
 
     const defaults = {
       cutoff: 5,
-      numberFormat: (d, value, total) => `${Number(d[value] / total * 100).toFixed(2)}%`,
+      numberFormat: (d, value, total) => {
+        const perc = Number(d[value] / total * 100);
+        return isNaN(perc) ? "No Data" : `${perc.toFixed(2)}%`;
+      },
       showText: "Show More",
       hideText: "Hide"
     };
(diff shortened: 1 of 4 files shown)


jhmullen:
@davelandry This is an interesting idea but how would the user pass in a function that depends on external variables?

`value` and `total` are variables that are only defined inside the scope of the rendering function of the PercentageBar.  If a user is authoring a `numberFormat` function in the config of a Viz, these variables are going to be undefined.

I could make the `numberFormat` super dumb, in that it really is just a formatter -- that is, it applies a text formatter to the raw number itself.  E.g. if d[value] is 17, the number format is just: 
```
d => `${d}%`
```

But this doesn't address my initial issue.  There are two fully different ways of displaying your data, one is to show the raw number (332 infections) and the other is to show its percentage (23%).  I agree that a unified function would be the way to handle this, but the __calculation__ of these figures depend on other elements of the config itself, such as `total` and `value`.  How should I tackle this?

jhmullen:
@davelandry You may disregard my earlier question, I've tackled this by making the arguments of numberFormat `(d, value, total)` so the user may define whatever formatting they would like.

packages/cms/src/components/Viz/PercentageBar.jsx
```diff
@@ -42,28 +44,30 @@ class PercentageBar extends Component {
 
     if (!config) return null;
 
-    const {data, cutoff, title, value, groupBy, total} = config;
+    const {data, cutoff, title, value, groupBy, total, numberFormat, showText, hideText} = config;
 
     const displayData = showAll ? data : data.slice(0, cutoff);
+
+    console.log(numberFormat);
```
davelandry: plz delete

packages/cms/src/components/Viz/PercentageBar.jsx
```diff
@@ -16,7 +15,10 @@ class PercentageBar extends Component {
     const {dataFormat} = this.props;
 
     const defaults = {
-      cutoff: 5
+      cutoff: 5,
+      numberFormat: (d, value, total) => `${Number(d[value] / total * 100).toFixed(2)}%`,
```
davelandry: catch for `isNaN` here?

jhmullen: I added the catch in both here and into the `reduce`, where I discovered that non-numerics can break my totaling logic.

packages/cms/src/components/Viz/PercentageBar.jsx
jhmullen: removed.

jhmullen:
@davelandry Please hold off on merging this.  I am going to backpack the work for #386 onto this PR.  Should be in by EOB today.

jhmullen:
@davelandry I have added in the extra functionality for #386 so with your approval this should be good to merge.

PR #330, opened 2018-12-06
Feature viz simple mode

jhmullen:
Adds a "simple mode" to visualizations (in addition to generators).  This feature (and the generator one as well) is still in need of "URL" replacement for things like `<id>` but this PR is to get the main UX flow in place for designing by @perpetualgrimace

### DIFF
--- a/packages/cms/src/api/cmsRoute.js
+++ b/packages/cms/src/api/cmsRoute.js
@@ -492,4 +492,15 @@ module.exports = function(app) {
     });
   });
 
+  app.get("/api/test", (req, res) => {
+    res.json([
+      {id: "alpha", x: 4, y: 7},
+      {id: "alpha", x: 5, y: 25},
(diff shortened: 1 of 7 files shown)


packages/cms/src/api/cmsRoute.js
```diff
@@ -492,4 +492,15 @@ module.exports = function(app) {
...
+    res.json({data: [
+      {id: "alpha", x: 4, y: 7},
+      {id: "alpha", x: 5, y: 25},
+      {id: "alpha", x: 6, y: 13},
+      {id: "beta",  x: 4, y: 17},
+      {id: "beta",  x: 5, y: 8},
+      {id: "beta",  x: 6, y: 13}
+    ]}).end();
+  });
+
```
davelandry: move this to `packages/cms/api/`

PR #369, opened 2018-12-19
[cms] Feature dev only

jhmullen:
Adds middleware to Create, Update, and Delete routes of CMS that restrict edit actions to development servers only.


jhmullen:
Issues above have been addressed @davelandry

PR #370, opened 2018-12-20
adds titles to footnotes

jhmullen:
Footnotes need titles for CDC.  Here they are!


davelandry:
thanks @jhmullen! can you go ahead and make these 2 new columns directly in the CDC database as well?

PR #405, opened 2019-01-15
[core] Bundle/code splitting

frabarz:
This enables basic bundle splitting for vendor code. More fine grained splitting can be achieved following documentation, but on my tests, webpack is smart enough.
Chunk injection will be required for future configurations of bundle and code splitting.

### DIFF
--- a/src/server.jsx
+++ b/src/server.jsx
@@ -73,19 +73,27 @@ const baseTag = process.env.CANON_BASE_URL === undefined ? ""
   : `
     <base href='${ BASE_URL }'>`;
 
-function getGeneratedTags() {
-  const filePath = path.join(process.cwd(), 'static/assets/index.html');
-  const html = fs.readFileSync(filePath).toString();
-  const headLink = html.match(/<link [^>]+>/g) || [];
-  const bodyScripts = html.match(/<script [^>]+><\/script>/g) || [];
-  return {headLink, bodyScripts};
+function getGeneratedTags(base) {
+  let headLink = [`<link rel="stylesheet" type="text/css" href="/assets/styles.css" />`];
+  let bodyScripts = [`<script type="text/javascript" src="/assets/app.js"></script>`];
+  if (process.env.NODE_ENV == "production") {
+    const filePath = path.join(process.cwd(), 'static/assets/index.html');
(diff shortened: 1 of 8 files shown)


src/server.jsx
```diff
@@ -70,10 +73,19 @@ const baseTag = process.env.CANON_BASE_URL === undefined ? ""
   : `
     <base href='${ BASE_URL }'>`;
 
+function getGeneratedTags() {
+  const filePath = path.join(process.cwd(), 'static/assets/index.html');
```
davelandry: this logic only works in prod. If I run `npm run dev` without having built the production build, I get an error because `index.html` has never been generated.

webpack/prod.js
```diff
@@ -22,6 +23,17 @@ module.exports = [
...
+      splitChunks: {
+        chunks: 'all',
+        cacheGroups: {
+          vendor: {
+            test: /[\\/]node_modules[\\/]/,
+            name: 'vendors'
+          }
+        }
+      }
+    },
```
davelandry: let's talk about `1.styles.css`: when looking at it's contents, it looks like it's only the blueprint styles... but those styles are also present in `styles.css`. What can we do about this duplication?

frabarz:
About the import procedure, I think is important to make the distinction: up this point, the PR only addresses _bundle splitting_, which is only chunking the vendors into smaller parts to make a better use of the browser's cache, and make loads faster when there's changes on the site's code.  This doesn't involve modifying the codebase, just the building phase.
On the other hand, _code splitting_ involves the use of `import()` in the codebase to tell webpack how it should split the codebase to reduce the size of the download needed to render the site. I'm still working on that, but as a spoiler, I got a way to make a higher-order component to generate the classes that will go on the `routes.jsx` file. It'll look like this:
```jsx
import {loadComponent} from "@datawheel/canon-core";
...
const ProfileLoader = loadComponent(() => import("path/to/Profile"));
...
<Route path="/profile/:id" component={ProfileLoader} />
```

src/server.jsx
frabarz: I'll address this on the next commit 👍

davelandry:
code splitting was merged with #1122

PR #419, opened 2019-01-28
changes subtitles from varchar to text to make room for more characters

### LINKED ISSUE #414: [cms] increase char limit of stat fields
We are using (kind of overloading) some of the stat fields with longer subtitles in CDC, and we've hit a 256 character limit. Can we change the type of these columns to be a longer text format?

jhmullen:
Closes #414 .


jhmullen:
FYI I have already made the appropriate changes in the live `dragonfly` db.

PR #429, opened 2019-02-08
[vizbuilder] general list on measure select, and feedback

### LINKED ISSUE #421: [vizbuilder] add ability to remove title
I would expect this to hide viz titles:

```js
config={{
  title: false
}}
```

### LINKED ISSUE #422: [vizbuilder] show full measure list when searching
When the user searches in the new nested measure selector, it can be a little confusing to have to click into each group that has a match to see what measure has matched. The idea here would be when search is triggered, the nesting mechanism goes away and shows results as they were before putting in the nesting.

### LINKED ISSUE #426: [vizbuilder] component does not render without defining measureConfig
Providing `measureConfig={{}}` gets past this, but is definitely a hack. Maybe a defaultProp is not set correctly?

### LINKED ISSUE #427: [vizbuilder] loading display should be controlled in React
currently, the display of the Loading component relies on the `fetching` class being present, and the class of the loading component being `loading`. I recently tried to put the vizbuilder into another project, and that project uses a different className for the Loading component. Ideally sites should be able to use whatever class they want, and the vizbuilder should show/hide the component based on a condition in React like this:

```jsx
{ load.inProgress && <LoadingScreen total={load.total} progress={load.done} /> }
```

frabarz:
closes #421 
closes #422 
closes #426 
closes #427

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -69,22 +69,23 @@ function moduleName(path) {
 /**
     Uses require.resolve to detect if a file is present.
 */
-function resolve(file) {
+function resolve(file, prefix = appPath) {
 
-  const fullPath = path.join(appPath, file);
+  const fullPath = path.join(prefix, file);
+  const dir = prefix === appPath ? ".app/ directory" : prefix === canonPath ? "canon-core source" : "custom directory";
 
   try {
     require.resolve(fullPath);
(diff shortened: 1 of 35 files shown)


davelandry:
👍 for work done
👎 for the name of the branch!

PR #489, opened 2019-03-15
Re-orders stat divs in Card

gtdane51:
Re-orders stat divs in Card for https://github.com/Datawheel/cdc-pha/issues/798

### DIFF
--- a/packages/cms/src/components/topics/Card.jsx
+++ b/packages/cms/src/components/topics/Card.jsx
@@ -15,18 +15,27 @@ export default class Card extends Component {
     const selectors = contents.selectors || [];
 
     const statGroups = nest().key(d => d.title).entries(stats);
+    const statGroupsMulti = nest().key(d => d.title).entries(stats);
+    statGroupsMulti.shift();
 
-    return <div className={ `topic pt-card pt-elevation-0 ${slug} Card` } ref={ comp => this.topic = comp }>
+    return <div className={ `topic bp3-card bp3-elevation-0 ${slug} Card` }>
       <div className="topic-content">
-        { title &&
-          <h3 className="topic-title">
-            <AnchorLink to={ slug } id={ slug } className="anchor" dangerouslySetInnerHTML={{__html: title}}></AnchorLink>
-          </h3>
-        }
+        <div className="title-wrapper">
+          { title &&
+            <h3 className="topic-title">
+              <AnchorLink to={ slug } id={ slug } className="anchor" dangerouslySetInnerHTML={{__html: title}}></AnchorLink>
+            </h3>
+          }
+
+          { stats.length > 0
+            ? <div className="topic-stats">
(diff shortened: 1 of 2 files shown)


packages/cms/src/components/topics/Card.jsx
```diff
@@ -15,18 +15,27 @@ export default class Card extends Component {
     const selectors = contents.selectors || [];
 
     const statGroups = nest().key(d => d.title).entries(stats);
+    const statGroupsMulti = nest().key(d => d.title).entries(stats);
+    statGroupsMulti.shift();
 
-    return <div className={ `topic pt-card pt-elevation-0 ${slug} Card` } ref={ comp => this.topic = comp }>
+    return <div className={ `topic bp3-card bp3-elevation-0 ${slug} Card` }>
```
davelandry: These `bp3` classes should stay as `pt`. You may have noticed an issue if you had installed canon-core v0.17, but cms v0.1 is locked to v0.16 of canon-core, which uses the `pt` prefixed.

packages/cms/src/components/topics/Card.jsx
```diff
@@ -15,18 +15,27 @@ export default class Card extends Component {
...
+          { stats.length > 0
+            ? <div className="topic-stats">
+              <StatGroup key={statGroups[0].key} title={statGroups[0].key} stats={statGroups[0].values} />
+            </div> : null }
+        </div>
         { subtitles.map((content, i) => <div key={i} className="topic-subtitle" dangerouslySetInnerHTML={{__html: content.subtitle}} />) }
         { stats.length > 0
           ? <div className="topic-stats">
-            { statGroups.map(({key, values}) => <StatGroup key={key} title={key} stats={values} />) }
+            { statGroupsMulti.map(({key, values}) => <StatGroup key={key} title={key} stats={values} />) }
```
davelandry: Instead of running the `nest` function an additional time, only to modify the resulting array to pop out the first entry, just use `stats.slice(1)` here to start at the second item. Additionally, you'll want to change the `stats.length > 0` at line 36 to `stats.length > 1`.

packages/cms/src/components/topics/Card.jsx
```diff
@@ -15,18 +15,26 @@ export default class Card extends Component {
     const selectors = contents.selectors || [];
 
     const statGroups = nest().key(d => d.title).entries(stats);
+    const statGroupsMulti = statGroups.slice(1);
 
-    return <div className={ `topic pt-card pt-elevation-0 ${slug} Card` } ref={ comp => this.topic = comp }>
+    return <div className={ `topic pt-card pt-elevation-0 ${slug} Card` }>
```
davelandry: Sorry I didn't catch this before, but can we keep this `ref` tag in there?

packages/cms/src/components/topics/Card.jsx
davelandry: I would just replace `statGroupsMulti` with `stats.slice(1)` here, no need to create a new variable.

PR #522, opened 2019-04-25
Refactors cms to await

### LINKED ISSUE #510: [cms] refactor all API routes to use async/await
The current structure might potentially be leading to memory leaks. Take a look at the refactored Data USA for inspiration: https://github.com/DataUSA/datausa-site/blob/canon/api/mortarRoute.js#L121-L275

I refactored all of the Data USA routes in about an hour, shouldn't take you too long here.

jhmullen:
Closes #510 by refactoring routes to await


jhmullen:
commit 1f7da7a addresses `verbose` and `console.error`.

Issue #23, opened 2017-08-02
Changing subdomains doesn't carry over user session


alexandersimoes:
While using canon locales and user management, if a user is logged in then changes subdomains their logged in session is lost. Is this by design? Is this a security issues... not sure.

davelandry:
closing in favor of #119

PR #524, opened 2019-05-01
Issue 523

### LINKED ISSUE #523: [cms] improve varSwap error message
two improvements to be made:
- [x] use `console.error` instead of `console.log` (this will print the message out to stderr, which is captured separately by PM2 for easier debugging)
- [x] print out the topic/item ID in the message (this may be trickier, but we need some way to identify which item is causing the error)

jhmullen:
Closes #523 by adding better error logging and putting messages behind a `verbose` flag

### DIFF
--- a/packages/cms/src/api/formatters.js
+++ b/packages/cms/src/api/formatters.js
@@ -18,7 +18,7 @@ module.exports = function(app) {
               r.logic = code;
             }
             catch (e) {
-              console.log("Error in Formatter Syntax: ", e.message);
+              console.log("Error in Formatter Syntax: ", e.message, `Formatter Name: ${r.name}`);
               r.logic = "return \"N/A\";";
             }
           }
--- a/packages/cms/src/api/formatters.js
+++ b/packages/cms/src/api/formatters.js
@@ -18,7 +18,8 @@ module.exports = function(app) {
               r.logic = code;
             }
             catch (e) {
-              console.log("Error in Formatter Syntax: ", e.message, `Formatter Name: ${r.name}`);
+              console.error(`Error in Formatter ${r.name}`);
+              console.error(`Error message: ${e.message}`);
               r.logic = "return \"N/A\";";
             }
           }
--- a/packages/cms/src/api/mortarRoute.js
+++ b/packages/cms/src/api/mortarRoute.js
@@ -99,7 +99,8 @@ const formatters4eval = (formatters, locale) => formatters.reduce((acc, f) => {
     acc[name] = FUNC.parse({logic: f.logic, vars: ["n"]}, acc, locale);
   }
   catch (e) {
-    if (verbose) console.error("Server-side Malformed Formatter encountered: ", e.message, `Formatter name: ${name}`);
+    console.error(`Server-side Malformed Formatter encountered: ${name}`);
+    console.error(`Error message: ${e.message}`);
     acc[name] = FUNC.parse({logic: "return \"N/A\";", vars: ["n"]}, acc, locale);
   }
 
--- a/packages/cms/src/api/mortarRoute.js
+++ b/packages/cms/src/api/mortarRoute.js
@@ -94,7 +94,7 @@ const formatters4eval = (formatters, locale) => formatters.reduce((acc, f) => {
     acc[name] = FUNC.parse({logic: f.logic, vars: ["n"]}, acc, locale);
   }
   catch (e) {
-    console.log("Server-side Malformed Formatter encountered: ", e.message);
+    console.log("Server-side Malformed Formatter encountered: ", e.message, `Formatter name: ${name}`);
     acc[name] = FUNC.parse({logic: "return \"N/A\";", vars: ["n"]}, acc, locale);
   }
 
--- a/packages/cms/src/components/Viz/index.jsx
+++ b/packages/cms/src/components/Viz/index.jsx
@@ -19,8 +19,6 @@ class Viz extends Component {
 
     const variables = this.props.variables || this.context.variables;
     const locale = this.props.locale || this.context.locale;
-    const id = this.props.id || "No ID Provided to Viz Component";
-
 
     // This Viz component may be embedded in two ways - as a VisualizationCard in the
     // CMS, or as an actual Viz on a front-end site. In the former case, formatters
@@ -31,6 +29,7 @@ class Viz extends Component {
     const formatters = this.context.formatters[locale] || this.context.formatters;
 
     const {config, configOverride, className, options, slug, topic} = this.props;
+    const {id} = config;
 
     // clone config object to allow manipulation
     const vizProps = propify(config.logic, formatters, variables, locale, id);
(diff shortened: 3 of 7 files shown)


packages/cms/src/utils/varSwapRecursive.js
```diff
@@ -46,7 +49,9 @@ const varSwapRecursive = (sourceObj, formatterFunctions, variables, query = {},
             obj[skey] = code;
           }
           catch (e) {
-            console.log("Error in ES5 transpiling in varSwapRecursive");
+            if (verbose) {
+              console.error(`Error in ES5 transpiling in varSwapRecursive.\nID was: ${obj.id}\nLogic attempted was: ${obj[skey]}`);
```
davelandry: - don't put this behind `verbose` (we will probably always want to show errors on prod servers)
- don't print the full logic (it's unnecessary if we have the ID and bloats stdout)
- make the message all on one line: `Error in ES5 transpiling in varSwapRecursive (ID ${obj.id})`
- is there any benefit in printing out `e.message` here?

packages/cms/src/utils/varSwap.js
```diff
@@ -27,7 +30,7 @@ module.exports = (sourceString, formatterFunctions, variables) => {
         return formatter(value, libs, formatterFunctions);
       }
       catch (e) {
-        console.log("Formatter error: ", e.message);
+        if (verbose) console.error("Formatter error: ", e.message, formatter);
```
davelandry: - don't put this behind `verbose` (we will probably always want to show errors on prod servers)
- let's do more explicit text, something like `Error using formatter "${formatter.name}" with variable ${value}`
- and then `e.message` on the next line?

packages/cms/src/utils/d3plusPropify.js
```diff
@@ -14,7 +14,7 @@ export default (logic, formatters = {}, variables = {}, locale = envLoc) => {
   }
   // If the javascript fails, return a special error object for the front-end to use.
   catch (e) {
-    console.log(`Parsing Error in propify: ${e}`);
+    console.error(`Parsing Error in propify: ${e}, Logic attempted was: ${logic}`);
```
davelandry: I actually added something similar to this into datausa-site, where I now pass the viz id as an additional argument:

```js
console.error(`Parsing Error in propify${id !== false ? ` (${id})` : ""}: ${e}`);
```

packages/cms/src/api/mortarRoute.js
```diff
@@ -99,7 +99,7 @@ const formatters4eval = (formatters, locale) => formatters.reduce((acc, f) => {
     acc[name] = FUNC.parse({logic: f.logic, vars: ["n"]}, acc, locale);
   }
   catch (e) {
-    if (verbose) console.error("Server-side Malformed Formatter encountered: ", e.message);
+    if (verbose) console.error("Server-side Malformed Formatter encountered: ", e.message, `Formatter name: ${name}`);
```
davelandry: - don't put this behind `verbose` (we will probably always want to show errors on prod servers)

packages/cms/src/api/formatters.js
```diff
@@ -18,7 +18,7 @@ module.exports = function(app) {
               r.logic = code;
             }
             catch (e) {
-              console.log("Error in Formatter Syntax: ", e.message);
+              console.log("Error in Formatter Syntax: ", e.message, `Formatter Name: ${r.name}`);
```
davelandry: `console.error`!

packages/cms/src/utils/varSwapRecursive.js
jhmullen: The reason I was printing the full logic was: 

- In case we use this transpiler in the future, perhaps for generic logic that isn't a visualization. Visualizations can be assumed to have an `id` but maybe some future case wouldn't.
- The id of a visualization is not present in the CMS at all. Seeing "error in id 3" would require the extra step of opening up your db directly to figure out which one I'm talking about (note that visualizations don't have plaintext names either).  Viewing the logic itself in the stacktrace was a way for me to give the user the "oh it's that one" helping hand.

Thoughts on that?

Regarding your second question, printing out `e.message` is indeed probably helpful here, because it would give more info on why the __transpiling__ failed, which is a different error from the OTHER error you will get later, which is why the es6 may have been malformed (handled in d3pluspropify)

davelandry: I understand, but I think that this is an opportunity to pass additional information to the function instead of printing out the logic (I'm worried about printing out like, ~50 line viz/generator logic here instead of just being able to say "hay it's viz #23"). I actually started passing through the viz ID to propify in datausa for this very reason: https://github.com/DataUSA/datausa-site/blob/master/app/components/Viz/index.jsx#L21

thoughts? still open to something other than that

jhmullen: This sounds good to me.

packages/cms/src/api/formatters.js
jhmullen: Fixed

packages/cms/src/utils/varSwap.js
jhmullen: Done

jhmullen:
@davelandry Latest push to the `issue-523` branch should address all this feedback.

packages/cms/src/components/Viz/index.jsx
```diff
@@ -19,6 +19,8 @@ class Viz extends Component {
 
     const variables = this.props.variables || this.context.variables;
     const locale = this.props.locale || this.context.locale;
+    const id = this.props.id || "No ID Provided to Viz Component";
```
davelandry: not exactly what I meant... should we not just use `config.id`?

packages/cms/src/utils/varSwapRecursive.js
```diff
@@ -1,6 +1,9 @@
 const selSwap = require("./selSwap");
 const varSwap = require("./varSwap");
 const buble = require("buble");
+const yn = require("yn");
+
+const verbose = yn(process.env.CANON_CMS_LOGGING);
```
davelandry: are these 2 variables needed anymore?

packages/cms/src/components/Viz/index.jsx
jhmullen: Sure we can move it into the config itself, but either way, the user would be responsible with loading up the config with this parameter.  It's easy for me to do this in the VisualizationCards of the CMS, but when people are making vizes on the front end, they will need to get the id from the original json payload, and inject it into the config object when they instantiate a `Viz`, otherwise this will come up null (Remember, the `logic` itself does not have an id in it).

Either way, having it as part of the config rather than a prop is cleaner - making that change.

packages/cms/src/components/cards/VisualizationCard.jsx
```diff
@@ -185,7 +186,7 @@ class VisualizationCard extends Component {
             onSave={this.save.bind(this)}
           />
         </Dialog>
-        { logic && !isOpen ? <Viz config={{logic}} locale={locale} variables={variables} configOverride={{height, scrollContainer: "#item-editor"}} options={false} /> : <p>No configuration defined.</p> }
+        { logic && !isOpen ? <Viz config={{id, logic}} locale={locale} variables={variables} configOverride={{height, scrollContainer: "#item-editor"}} options={false} /> : <p>No configuration defined.</p> }
```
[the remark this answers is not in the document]
jhmullen: Agree - Latest push passes entire object into the config

PR #527, opened 2019-05-07
fixes bug in varSwapRecursive that caused levels to improperly turn into objects

jhmullen:
in varSwapRecursive, I was making the assumption that if the object i've encountered is an array, it is sufficient to recursively run varSwapRecursive on it again.

However, profiles now have `levels`, which is simply an array of strings. Therefore, I was trying to run varSwapRecursive on a __string__, which was casting it to an object and breaking the datatype.

I added a case into the "array" part of the function.  As it fans over the array, if it is an object, we use varSwapRecursive, but if it's a string, we have "hit bottom" and can use the new `strSwap`.

Fixes a bug that @cnavarreteliz discovered in Estonia with hierarchies.

### DIFF
--- a/packages/cms/src/utils/varSwapRecursive.js
+++ b/packages/cms/src/utils/varSwapRecursive.js
@@ -2,6 +2,28 @@ const selSwap = require("./selSwap");
 const varSwap = require("./varSwap");
 const buble = require("buble");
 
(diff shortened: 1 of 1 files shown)


jhmullen:
@davelandry latest commit pulls in `master`.

PR #554, opened 2019-06-12
Adds new url-loader file extension

MarcioPorto:
`.JPG` files are not currently supported.

### DIFF
--- a/webpack/config/loaders.js
+++ b/webpack/config/loaders.js
@@ -57,7 +57,7 @@ module.exports = props => {
       ]
     },
     {
-      test: /\.(png|jpeg|jpg|gif|bmp|tif|tiff|svg|woff|woff2|eot|ttf)$/,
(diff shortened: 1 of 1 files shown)


webpack/config/loaders.js
```diff
@@ -57,7 +57,7 @@ module.exports = props => {
       ]
     },
     {
-      test: /\.(png|jpeg|jpg|gif|bmp|tif|tiff|svg|woff|woff2|eot|ttf)$/,
+      test: /\.(png|jpeg|jpg|gif|bmp|tif|tiff|svg|woff|woff2|eot|ttf|JPG)$/,
```
davelandry: I believe the `i` flag at the end of a regex pattern makes it case insensitive, so theoretically this may be a more robust match pattern:

```js
/\.(png|jpeg|jpg|gif|bmp|tif|tiff|svg|woff|woff2|eot|ttf)$/i
```

@MarcioPorto can you test that out and see if it still works? If so, let's use that instead.

PR #581, opened 2019-07-05
Feature sticky sections

### LINKED ISSUE #577: [cms] Implement Front-end support for sticky sections
CSS/JS in abstract `Section` class (@perpetualgrimace):   - sticks to top of page when it hits it   - moves off of top when the next sticky section replaces it

perpetualgrimace:
closes #577 

- Adds support for sticky profile sections.
- Adds sticky section fallback in IE, which is a homebrew fallback using refs, IE detection, and throttling. It also offsets the top of the screen by using `styles["nav-height"]`. 
- *Note*: I was unable to actually test the fallback in IE, since currently IE won't load _any javascript at all_ due to the `"use strict";` error. I've removed the polyfill libraries I was testing, cleared the node modules cache, reran, and still got the error in IE. So I'm assuming this is a separate issue.


perpetualgrimace:
😭😭😭

Okay, I'll dig back into the IE thing soon.

As for #2, that's a good point. It's probably doable, but that's not `position: sticky` works. What it does is treats an element as `position: relative` until it hits the offset specified (i.e., `top: 0`), then it switches to `position: fixed`. And vice versa. I'm worried about managing even more the complexity with scroll listeners.

I'm already worried about like, what if the navbar isn't fixed? What if the subnav we eventually (re)implement is or isn't fixed? What if things are fixed, or not, at different viewports (which they should be). What if there's another element altogether that needs to be fixed. 🤔

I think this is a feature that can really be problematic. I think we should explicitly set a height (can be defined as a css variable) so that each new sticky section is fully hidden by the next and it's never janky. That would require making some assumptions about the layout. Perhaps what we need is a...wait for it...sticky section (kills self)

davelandry:
@perpetualgrimace totally get your points. Let's not worry about the sticky section overlap for now.

As for the offset... could we create something like `--sticky-offset`, and in the default values we fallback to use `--nav-height`?

perpetualgrimace:
Oh that's smart. Yes. Good idea.

PR #586, opened 2019-07-12
[cms] includes new d3plus-react formatters to Visualizations

cnavarreteliz:


packages/cms/src/components/Viz/index.jsx
```diff
@@ -67,6 +67,9 @@ class Viz extends Component {
         ref={ comp => this.viz = comp }
         className="d3plus"
         dataFormat={resp => (this.analyzeData.bind(this)(resp), vizProps.dataFormat(resp))}
+        linksFormat={resp => (this.analyzeData.bind(this)(resp), vizProps.linksFormat(resp))}
+        nodesFormat={resp => (this.analyzeData.bind(this)(resp), vizProps.nodesFormat(resp))}
+        topojsonFormat={resp => (this.analyzeData.bind(this)(resp), vizProps.topojsonFormat(resp))}
```
davelandry: This won't work, because the fallback for these functions (unlike the `dataFormat`) is `undefined`. The `analyzeData` function is what grabs the data source and adds that text underneath the paragraphs. We don't need that for these files, so just pass the values:

```jsx
topojsonFormat={vizProps.topojsonFormat}
```

davelandry: @cnavarreteliz why are you still wrapping these functions in an anonymous pass-through function? Can't we just do `topojsonFormat={vizProps.topojsonFormat}` and save that extra function step?

PR #587, opened 2019-07-12
Migration Scripts

### LINKED ISSUE #576: [cms] Create Update Scripts for various old CMS versions
- [x] Data USA update script
- [x] v0.1 (CDC) update script
- [x] v0.6 update script

jhmullen:
This branch includes three new migration scripts:

1) `npm run migrate-legacy` (for DataUSA) 
2) `npm run migrate-0.1` (for CDC or other 0.1 CMS users)
3) `npm run migrate-0.6` (for 0.6 CMS users)

### Instructions

The name of the script represents the version you wish to migrate **FROM**.  So, to upgrade a DB from 0.6 to 0.7, one would use `npm run migrate-0.6`.  Currently both `legacy` and `0.1` upgrade directly to `0.6`.  From here on out, versions will upgrade **one dot version at a time**.

It is necessary that users spin up an entire new database for any CMS migration.

The user need configure two sets of environment variables, `OLD` and `NEW`.

```
CANON_CONST_MIGRATION_OLD_DB_NAME
CANON_CONST_MIGRATION_OLD_DB_USER
CANON_CONST_MIGRATION_OLD_DB_PW 
CANON_CONST_MIGRATION_OLD_DB_HOST

CANON_CONST_MIGRATION_NEW_DB_NAME
CANON_CONST_MIGRATION_NEW_DB_USER
CANON_CONST_MIGRATION_NEW_DB_PW
CANON_CONST_MIGRATION_NEW_DB_HOST
```

These variables represent the old db you are migration **from** and the new db you are migrating **to**.  The new db will be **wiped every time** you run the script - the idea here is that you are building a new db from scratch.

🔥 WHATEVER DB YOU CONFIGURE AS **NEW** WILL BE COMPLETELY DESTROYED AND BUILT FROM SCRATCH 🔥
🔥 DO NOT SET `CANON_CONST_MIGRATION_NEW_DB_*` TO A CURRENTLY IMPORTANT DB🔥

After the migration is done, you can switch your dev environment to the new DB for testing, and eventually switch it to prod.

Closes #576


packages/cms/package.json
```diff
@@ -11,7 +11,9 @@
     "release": "canon-release",
     "setup": "canon-setup",
     "start": "canon-start",
-    "migrate-db": "node scripts/migrate-db.js"
+    "migrate-legacy": "node scripts/migration/migrate-legacy.js",
+    "migrate-0.1": "node scripts/migration/migrate-0.1.js",
+    "migrate-0.6": "node scripts/migration/migrate-0.6.js"
```
davelandry: _cracks knuckles_ 🤓 

Let's get these scripts exported from this package. Adding these to the `package.json` "scripts" object works locally when in this folder, but do not "ship" with the package. For that, we need to add them to the "bin" object. Check out how `canon-core` is setup: https://github.com/Datawheel/canon/blob/master/package.json#L6-L12

Additionally, let's give the commands a little more namespace. How about `canon-cms-migrate-*`?

Aaaaaand finally, as an FYI, projects can run a package's script by typing `npx script-name` instead of having to actually add it to the package.json "scripts" object, which is great for one-time scripts like these (so that's what we should use in usage documentation).

davelandry: @jhmullen and ANOTHER thing: at some point ago, to get the "bin" scripts picked up for local testing, it was necessary to also do the following (YMMV):
* Add [this one-liner](https://github.com/Datawheel/canon/blob/master/bin/dev.js#L1) to the top of any main script file: `#! /usr/bin/env node`
* After you add the scripts to package.json, you needed to run `npm run link` to "register" them in your local env. This script gets run automatically when installing the package in another repo.

packages/cms/README.md
```diff
@@ -37,3 +37,40 @@ import {Profile} from "@datawheel/canon-cms";
...
 |---|---|---|
 |`CANON_CMS_ENABLE`|Setting this env var to `true` allows access to the cms in production builds.|`false`|
+
+---
+
+## Migration
+
+For upgrading to new versions, there are currently three migration scripts:
+
+1) `npm run migrate-legacy` (for DataUSA)
```
davelandry: Shouldn't these all be `npx canon-cms-migrate-legacy`?

packages/cms/package.json
jhmullen: @davelandry added readme with `npx`, and namespaced the scripts!

packages/cms/README.md
```diff
@@ -37,3 +37,40 @@ import {Profile} from "@datawheel/canon-cms";
...
+
+---
+
+## Migration
+
+For upgrading to new versions, there are currently three migration scripts:
+
+1) `npx migrate-legacy` (for DataUSA) 
+2) `npx migrate-0.1` (for CDC or other 0.1 CMS users)
+3) `npx migrate-0.6` (for 0.6 CMS users)
```
davelandry: @jhmullen those aren't the names of the scripts 😑 

(missing the `canon-cms-` prefix)

jhmullen: Good catch, My bad, fixed.

PR #588, opened 2019-07-15
redux entire custom middleware wip

palamago:
My idea about custom middleware.
It is working BUT: would it will work with a middleware exported from third party libs?


palamago:
Here is the example from Kepler: https://github.com/keplergl/kepler.gl#1-mount-reducer

src/storeConfig.js
```diff
@@ -40,14 +47,13 @@ export default function storeConfig(initialState, history) {
   }, appReducers));
 
   if (__DEV__ && !__SERVER__) {
-    if (__LOGREDUX__) middleware.push(createLogger());
     store = createStore(canonReducer, initialState, compose(
-      applyMiddleware(...middleware),
+      appliedMiddleware,
       typeof window === "object" && typeof window.devToolsExtension !== "undefined" ? window.devToolsExtension() : f => f
```
frabarz: the exposed function by redux devtools changed; this property will need to be updated in the future
https://github.com/zalmoxisus/redux-devtools-extension#usage

davelandry: @frabarz just pushed a tweak, what do you think?

PR #610, opened 2019-07-25
fixes bug in viz assign

jhmullen:
A bug in config assignment breaks when multiple vizes are on a page, overloading the `defaultConfig` object and breaking vizes.  Assigning it to an empty object fixes this.


davelandry:
`git-blame davelandry` 😅

PR #628, opened 2019-08-01
Feature table viz

perpetualgrimace:
371mix


packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -102,7 +105,12 @@ class SimpleVisualizationEditor extends Component {
           return `\n  "${k}": \`${fixedUrl}\``;
         }
         else {
-          return `\n  "${k}": "${object[k]}"`;
+          if (k === "columns") {
+            return `\n "${k}": [${object[k].map(d => `"${d}"`).join()}]`;
```
jhmullen: Could you please change my `join()` hackjob into the `JSON.stringify()` that was suggested today? 
```suggestion
            return `\n "${k}": ${JSON.stringify(object[k])}`;
```

packages/cms/src/utils/formatters/abbreviate.js
```diff
@@ -1,4 +1,4 @@
-const libs = require("./libs");
+const libs = require("../libs");
```
jhmullen: If this required correcting in `abbreviate.js` and `commas.js`, does it require fixing in other places? Could you please review all other functions in `/formatters` and adjust accordingly?

PR #630, opened 2019-08-02
hide visualization titles

### LINKED ISSUE #619: [cms] hide visualization titles unless a `title`  is explicitly defined in the viz config

perpetualgrimace:
Hides visualization titles unless a `title` is explicitly defined in the viz config; closes #619

### DIFF
--- a/packages/cms/src/components/Viz/Viz.jsx
+++ b/packages/cms/src/components/Viz/Viz.jsx
@@ -116,7 +116,6 @@ Viz.defaultProps = {
   configOverride: {},
   context: "cp",
   options: true,
-  title: undefined,
   showTitle: true,
   headingLevel: "h3"
 };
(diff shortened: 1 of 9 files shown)


packages/cms/src/components/Viz/Viz.jsx
```diff
@@ -55,7 +55,7 @@ class Viz extends Component {
       return <div>{`${type} is not a valid Visualization Type`}</div>;
     }
 
-    const title = vizProps.config.title || this.props.title || config.title || slug || "";
+    const title = vizProps.config.title;
```
jhmullen: You made a change two weeks ago to this line that needs to be revisited.

https://github.com/Datawheel/canon/commit/4cef9394643550de1d36bc3e2dccdf8c3ba78699#diff-1d8b00dab1ae8869fc8640e3d18814abR56

Changing the order of these vars changes the behavior - I want whoever implements the Viz component to be able to override the title using a title prop.

`vizProps.config.title || this.props.title`

This way we attempt to take the title in the native config, and if it's not there, we fall back on the title provided to the Viz via props, otherwise just give up an hide it (as you've already coded below).

Issue #611, opened 2019-07-26
[vizbuilder] adding a drilldown with a single member results in duplicate charts


davelandry:
check out this URL: /visualize?groups=0-1mjmRl-5112~51913&groups=1-1vG6TK-24&groups=2-1kzwvK&groups=3-2pRcAs&measure=1qWfo

If you remove the "Nation" drilldown, the duplicate chart goes away! Do we already have a local instance of the member list to check the number of members?

davelandry:
@frabarz this fixed seemed to have regressed. Using the latest vizbuilder I'm now not only seeing duplicate tree maps (I now see 12!), but we've also lost the nativity groupings:

<img width="1377" alt="Screen Shot 2019-08-02 at 11 13 02 AM" src="https://user-images.githubusercontent.com/1847581/62380186-b01b5500-b516-11e9-8f29-e13ee9c55eea.png">

PR #632, opened 2019-08-02
[vizbuilder] removes treemap groupings with 1 member

frabarz:
fixes treemap labels
related to #611


davelandry:
@frabarz looks great! can you also port this to v0.2? 😓

PR #680, opened 2019-09-06
PercentageBar Updates

rhauckdot:
Added SVG ticks and labels to PercentageBar. Added isPercentValue boolean parameter to determine type of labels to display.

### DIFF
--- a/packages/cms/src/components/Viz/PercentageBar.css
+++ b/packages/cms/src/components/Viz/PercentageBar.css
@@ -2,6 +2,7 @@
 
 .percentage-bar-list {
   @mixin list-reset;
+  padding-bottom: 0;
 }
 
 .percentage-bar-bg,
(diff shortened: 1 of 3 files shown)


packages/cms/src/components/Viz/PercentageBar.jsx
```diff
@@ -55,49 +53,83 @@ class PercentageBar extends Component {
...
+        labelVal = i === 0 ? 0 : (total*(i*0.1)).toFixed(2);
+      }
+
+      lines.push(<line key={i} x1={xPos} x2={xPos} y1="0" y2="100%"/>);
+      ticks.push(<text key={i} x={xPos} y="100%">{labelVal}</text>);
+    }
+
     return (
-      <div className="percentage-bar-wrapper">
+      <>
```
davelandry: what's this empty element that's been added?

perpetualgrimace: that's the shorthand for `<React.Fragment>`

packages/cms/src/components/Viz/PercentageBar.css
```diff
@@ -41,3 +42,88 @@
...
+    position: absolute;
+    width: 100%;
+    height:100%;
+}
+.percentage-bar-ticks{
+    clip-path: url(#percentage-bar-clip);
+}
+.percentage-bar-ticks line{
+    stroke: #e5e5e5;
+    stroke-width: 2;
```
perpetualgrimace: Please replace with `stroke-width: var(--percentage-bar-stroke-width)` and [add a new line to style.yml](https://github.com/Datawheel/canon/blob/master/packages/cms/app/style.yml#L209): `percentage-bar-stroke-width: "2px"`

perpetualgrimace:
Appreciate it! And don't worry — I'm still learning the ropes too. 😁

PR #699, opened 2019-09-18
Ap 10

rhauckdot:
Added styling to InfoCard to reflect revised design.


davelandry:
@perpetualgrimace these changes fall into your wheelhouse, so please give this PR a review, keeping an eye on 2 things:
1. Rich added a few more style colors, and wasn't sure if that's ok (or if he should have re-used others)
2. There are a few small changes to the abstract Section class, so make sure those don't break any other magic you put in there.

rhauckdot:
@perpetualgrimace Honestly, I have not tested the card in multiple columns (isn't this component specific to our project?). In Canon, cards are displayed in a 3-column layout (within `.cp-info-card-section-grouping`), whereas in our implementation it only permits two columns. In the 3-column layout the card's internal columns are squished...I would usually define a parent > child selector where I would format the card to the column width, but that hasn't been done here and I don't want to overstep my bounds.

I am trying to be respectful of existing props in components and therefore haven't renamed any props in `Section.jsx`. Previously, `heading` delivered the title and subtitle together, but the new card design requires them in separate areas. I define these separately as `mainTitle` and `subTitle`, but have them still returned together as `heading` (https://github.com/Datawheel/canon/pull/699/files#diff-cd0be3f90ec5ce142e2b5e41a8e310ebR148-R151).

davelandry:
@rhauckdot there are also a few linting errors in `InfoCard.jsx`(mainly unused variables). Can you install an eslint plugin into your text editor so that you can see the errors? (theres an `.eslintrc` file in the root directory that declares the rules)

perpetualgrimace:
@rhauckdot Hey! Apologies for my misunderstanding of the changes you made to `Section.jsx`. As long as you're not affecting the other sections we should be good.

So in general, when adding styles to canon components, it's important to make sure it holds up in different configurations and screen sizes, with different data. Trying to break a component is the best way to ensure it's flexible. You can bet users will break it unintentionally if you don't try first. 🙂

Last time I checked on the CDC PSP (I'll call it CDC from now on since that's how we refer to it internally), there is one profile type which shows a single info card by itself, and another profile type that groups them together in columns of three. But even if you could guarantee it will only be shown in columns of three, the project could change its requirements, and you can bet people will view them on mobile. 

The responsive layout stuff should be largely taken care of out of the box, but I just want to make sure you're testing for that.

rhauckdot:
@perpetualgrimace no worries. CDC-PHA, CDC-PSP...I think I'll join you with just "CDC". 

I'm not concerned with the InfoCard layout at desktop, but at desktop within 3 columns it gets a bit squished. A media query can't be used here since it's the columns compressing and not the browser width. Besides that, I did place a media query for < 767px width. If I should modify this further let me know.

rhauckdot:
@davelandry thanks (I thought I had already installed ESLint in VSCode). I removed nest, `stripP`, `StatGroup`, and `Selector` imports. Coincidentally, the only remaining unused value is `heading`.

I consolidated `.cp-info-card-section-heading` styling into `.cp-info-card-section-header`.

I'm still not seeing values from style.yml resolve locally. `.cp-info-card-section-header`, for instance, uses `background-color: var(--infocard-header-bg-color);`. When I view this in the browser inspector it lists it as `background-color: var(--light-1);`, which is really odd, as my style.yml is set to `var(--dark-3)` (https://github.com/Datawheel/canon/pull/699/files#diff-01159378f9ae968cbc3c0e15f3dfa483R166).

rhauckdot:
@davelandry I feel dumb for asking, but what package should I call that npm command? Will it be corrected in a future release?

The infocard title ("CLABSI") and the select labels in the header have attributes of `.cp-info-card-section-header` and `.cp-select-label.cp-select-text`, respectively. Both of these should be referencing `var(--infocard-header-text-color)` from style.yml. That should be set to `--white`, though I'm wondering if it's being overrided. If I can get a local instance running I'll investigate.

rhauckdot:
I pulled https://github.com/Datawheel/canon/tree/cms-0.7 into my fork ap-10 branch and it wouldn't populate the profiles. I suspect it's because  redux-logger is reporting this: `{ type: "LOGIN_FAILURE", payload: Error }`

My local branch of cms-0.7 does connect, so I branched that and tried copying ap-10's changes over. The result is a similar to what I saw before.

Screen Shot 2019-09-27 at 9 23 18 AM

InfoCard title is missing, key data points `{secondaryStats}` isn't returning anything, and `{stats}` is returning all stats instead of only the first.

davelandry:
@rhauckdot ok so, I was able to get this working. To do so, I ported the `cms-0.7` to `0.8` for testing purposes, please change this environment variable (earwig still uses 0.7, and this is only for testing purposes):

```
export CANON_DB_NAME=cms-0.8
```

That way, this PR can still be merged with master, and when it is, we'll (for real) upgrade cdc-pha to cms 0.8.

Also, please make sure that your `CANON_API` env var is set to `http://localhost:3300`. This variable is responsible for all of the data calls to the CMS (and not cubes), so we want to make sure it's using the local 0.8 api, and not earwig's 0.7 api.

After I changed those 2 env vars, I was able to run dev on this branch and see the cards looking like the screenshot I previously posted.

Additionally, you don't need to worry about the "LOGIN_FAILURE" messages. Those have actually always been there (you didn't notice 😉) because the site always checks to see if there are any previously logged in user cookies (used in canon sites with user logins). Please ignore those messages in dev.

rhauckdot:
@davelandry I set my globals to:

```
export CANON_DB_NAME=cms-0.8
export CANON_API="http://localhost:3300"
```
I pushed another styling tweak, but I'm still seeing the stats rendering together in the header. Open to suggestions.

davelandry:
@rhauckdot which URL are you previewing? the only next thing I can think of is to delete your whole local folder and do a fresh git pull from this branch.

rhauckdot:
http://localhost:3300/profile/infections/vae/

The geography profile looks pretty good... 
http://localhost:3300/profile/geography/42

davelandry:
@rhauckdot looking at how you're rendering the stats in `InfoCard.jsx`, I see that it's using `stats` and `secondaryStats`. Tracing that back to `Section.jsx` where those variables are created, I see that it's using `statGroups`, which groups stats based on similar "labels".

Add some `console.log` statements in there to see what it's doing. My guess from a quick look is that the content in the CMS for the infections profile is formatted differently than the geography profile, and the infections profile can only determine 1 grouping while it's doing separate groupings for the geography profile as there is 1 stat there with a different title than the rest.

Either change the way the code groups the stats for InfoCards, or edit the CMS content to fit your needs.

rhauckdot:
@davelandry once I realized the geography InfoCards were working I immediately questioned if it was the entry causing the difference. So I added new stats, copied the existing stats into them, then deleted the old stats and...it's working. I don't believe there's any difference in the entries, leading me to wonder what flag defines inclusion into `statGroups`.

I still want to see CDC's styles.yml applied once it's been updated, otherwise, were there any additional revisions on your end?

perpetualgrimace:
You can (and should) override the style variables to your liking in the CDC project's style.yml file — we've just been setting canon defaults 🙂

davelandry:
@rhauckdot 🎉 looks great, I just merged this and released it as `@datawheel/canon-cms@0.8.10`. Upgrade the package in cdc-pha and you should see your changes.

perpetualgrimace:
Hey @rhauckdot,

My fault for not catching this, but you set a few styles in InfoCard.css that leaked out into the entire application and make some select menu labels invisible. 😬

Namely:
```
.cp-select-label {
  display: block;
  text-transform: uppercase;
  flex-direction: column;

  & .cp-select-text {
    color: var(--infocard-header-text-color);
    padding-bottom: var(--gutter-xs);
  }
}
```

and

```
.cp-stat-subtitle {
  color: inherit !important;
}
```

I've gone ahead and properly scoped the declarations to `.cp-info-card-section .cp-select-label { ... }` and `.cp-info-card-section .cp-stat-subtitle { ... }`, but please be careful moving forward.

PR #703, opened 2019-09-20
Canon CMS 0.8

### LINKED ISSUE #471: [cms] multilingual search
Right now we default to english when we populate the search table.  

Long term:

- [x] split the search table into two tables - actual items and language content
- [x] refactor the search endpoint to have a `?lang=` flag which filters content and returns ids

I have moved some QoL tasks to #638 so that a multi-lang MVP can get out the door. For now, when a user adds a new language, he or she will have to rebuild the search index manually in the CMS interface.

### LINKED ISSUE #636: [cms] implement flickr image management
Current Data USA google sheet: https://docs.google.com/spreadsheets/d/1emEA-Tz4EBugidBRSQ5e4eq9AkQaEqAerTW0XtGL4AI/edit#gid=1064521817

And node script: https://github.com/DataUSA/datausa-site/blob/staging/scripts/flickr-scrape.js

### LINKED ISSUE #642: [cms] show preview name in dimension search
I've seen a few people (myself included) get confused when trying to change the preview dimension. The input box is labeled as "preview ID" and shows the ID, but you cannot search by ID (only by name).

Maybe separate "preview ID" out to it's own separate line and display the name in the search box?

### LINKED ISSUE #659: write 0.8 migration script

### LINKED ISSUE #695: [cms] allow visualizations to edit state
Open for discussion, but my initial though is: what if we (somehow) provided the `setState` function to viz configs, and that would allow them to set custom properties on state.

For example, an click event in a Treemap:
```js
on: {
  click: d => setState({highlight: d.id})
}
```

And then, if they desire, other visualizations can access this value to change their contents:

```js
data: state.highlight ? `custom-url-${state.highlight}` : "default-url"
```

### LINKED ISSUE #697: [cms] Inject image into Hero
In the `feature-multisearch`, images uploads are working in the member builder, but we need to give easy access to that image in the Hero section.  Maybe provide it by default in the hero, but at least document how to access the images

### LINKED ISSUE #698: [cms] in meta editor, when an image is changed, the image doesn't update in the table

### LINKED ISSUE #700: [cms] Profiles broken when they have only one section
There is logic in `Profile.jsx` which was written to nest InfoCards and SingleColumns:

```
    // reduce sections into a nested array of groupedSections
    innerGroupedSections.push(sections.reduce((arr, section) => {
      if (arr.length === 0) arr.push(section); // push the first one
      else {
        const prevType = arr[arr.length - 1].type;
        const currType = section.type;
        // if the current and previous types are groupable and the same type, group them into an array
        if (groupableSections.includes(prevType) && groupableSections.includes(currType) && prevType === currType) {
          arr.push(section);
        }
        // otherwise, push the section as-is
        else {
          innerGroupedSections.push(arr);
          arr = [section];
        }
      }
      return arr;
    }, []));

    const groupedSections = innerGroupedSections.reduce((arr, group) => {
      if (arr.length === 0 || group[0].type === "Grouping") arr.push([group]);
      else arr[arr.length - 1].push(group);
      return arr;
    }, []);
```

However, when a user creates a new Profile, it only has one section, which defaults to a Hero Type.  If the user goes and tries to preview his/her brand new profile it breaks with the following error:

```
{
error: "TypeError: Cannot read property 'type' of undefined",
stackTrace: "TypeError: Cannot read property 'type' of undefined at eval (webpack:////Users/jimmy/Datawheel/canon/packages/cms/src/components/Profile.jsx?:231:38) at Array.map (<anonymous>) at eval (webpack:////Users/jimmy/Datawheel/canon/packages/cms/src/components/Profile.jsx?:223:25) at Array.map (<anonymous>) at Profile.render (webpack:////Users/jimmy/Datawheel/canon/packages/cms/src/components/Profile.jsx?:220:34) at processChild (webpack:////Users/jimmy/Datawheel/canon/packages/cms/node_modules/react-dom/cjs/react-dom-server.node.development.js?:3035:22) at resolve (webpack:////Users/jimmy/Datawheel/canon/packages/cms/node_modules/react-dom/cjs/react-dom-server.node.development.js?:2881:9) at ReactDOMServerRenderer.render (webpack:////Users/jimmy/Datawheel/canon/packages/cms/node_modules/react-dom/cjs/react-dom-server.node.development.js?:3291:26) at ReactDOMServerRenderer.read (webpack:////Users/jimmy/Datawheel/canon/packages/cms/node_modules/react-dom/cjs/react-dom-server.node.development.js?:3243:33) at renderToString (webpack:////Users/jimmy/Datawheel/canon/packages/cms/node_modules/react-dom/cjs/react-dom-server.node.development.js?:3775:31) at eval (webpack:///./server.jsx?:104:102) at processTicksAndRejections (internal/process/task_queues.js:89:5)"
}
```

I believe this is because of the `arr[arr.length - 1]` code, which will access negative indexes in the case of a zero-length array.  Please have another look at how you calculate the grouping to support the "only have a hero so far" rendering. 

**NOTE** Please make your changes in `master`.

### LINKED ISSUE #701: [cms] Make use of image data in Hero Component
In the `feature-multisearch` branch, I have added a new `images` object to state in Hero.jsx which contains a strictly ordered list of images for the slug/id pairs. 

Please add the image source itself to the background of the Hero (like in DataUSA) and also find places to put the `author`, `meta`, and `permalink` information. 

### More information

Note that unlike in DataUSA, we now have multi-variate profiles. This means that a given page could have 1, 2, or 3 entities in it, perhaps something like `/geo/mass/export/coal/import/cars`. You will need to find a way to show all three of these images (split evenly perhaps?).  However, there will also be cases where _not all entities have images_. 

As stated earlier, the list is strictly ordered.  So if `/export/coal/` is the only one with an image from the example above, `images` will look like `[null, {imageData}, null]`.

jhmullen:
# Canon CMS 0.8 Release 

### Multilingual Search

Canon CMS 0.8 makes use of the new [olap-client](https://www.npmjs.com/package/@datawheel/olap-client) to handle both mondrian and tesseract connections. If your cube has language annotations, the CMS will automatically import different languages into the search table based on your `CANON_LANGUAGES` env var.  The `/api/search` endpoint has been changed accordingly, allowing for a `?locale=<locale>` parameter to be set to restrict your searches by language.

### Flickr Image Management

The new `Meta Editor` tab of the CMS allows users to choose flickr images for search members, either by direct link or by searching for a keyword. In both cases, images with restrictive licenses will be rejected. All images are hosted on Google Cloud Storage. Many many thanks to @perpetualgrimace for doing the lion's share of the UX here.

See the [README](https://github.com/Datawheel/canon/blob/master/packages/cms/README.md#enabling-image-support) for information on how to configure google cloud storage and image management.

Images will automatically be rendered in the "Hero" section of a profile, which is automatically created upon profile creation. However, if you need direct access to the hosted images, they are reachable via:

```
/api/image?slug=<slug>&id=<id>
```

Images default to splash size, but you may set `&size=thumb` for a thumbnail. To retrieve metadata about the image rather than the image itself, add `&type=json` to the params.

### Member Management

The `Meta Editor` tab also allows users to set a comma-separated list of keywords for a given search member (e.g., `mit,cambridge` for Massachusetts Institute of Technology). Additionally, one may set "language hints," a javascript lookup object to assist in the rendering of language-specific rules on the front-end.

### Visualization Variable Override

Javascript in vizes now has the ability to call a `setVariables` function, which will inject whatever key/value pairs you provide into the overall `variables` object of the Profile page. This useful if you want visualization click events to cause other events to happen on a profile. For example, if you have a Visualization with a set of products, and you want to be able to click a product and spawn a new visualization based on that product:

1) Create a variable called `showSecondaryViz` in a Materializer, and set it to `false`.
2) In the onClick for your first viz, add something like:  `d => setVariables({secondaryID: d.id, showSecondaryViz: true})`
3) In the configuration for your second viz, set it's "Visible" property to `showSecondaryViz`, and make use of `variables.secondaryID` in your viz configuration.

### Migration

As usual, migrating to this new version will require a database migration script. Instructions are in the [README](https://github.com/Datawheel/canon/blob/master/packages/cms/README.md#enabling-image-support).

Closes #701
Closes #700
Closes #698 
Closes #697
Closes #659 
Closes #695
Closes #642
Closes #636 
Closes #471


packages/cms/src/components/DimensionCreator.jsx
```diff
@@ -78,11 +78,10 @@ class DimensionCreator extends Component {
         message: "Slug already taken. Please choose a different slug.",
         timeout: 2000
       });
-    }
+    }
```
davelandry: LOL REMOVE THIS WHITESPACE

packages/cms/src/components/Viz/Options.jsx
```diff
@@ -103,7 +103,7 @@ class Options extends Component {
       let background;
       if (backgroundColor) background = getBackground(node);
       saveElement(node,
-        {filename: filename(title), type},
+        {filename: `${filename(title)}.${type}`},
```
davelandry: I believe this may have unintended consequences. The `saveElement` function expects 2 keys, "filename" and "type", and without specifying type this may cause them all to use the fallback type ("PNG"), and would then name the files like this `my-title.png.png`.

packages/cms/src/components/Viz/Viz.jsx
```diff
@@ -20,6 +20,10 @@ class Viz extends Component {
     if (updateSource && resp.source) updateSource(resp.source);
   }
 
+  setVariables(newVariables) {
+    if (this.props.onSetVariables) this.props.onSetVariables(newVariables);
```
davelandry: Why do we have this pass-through function (don't all `Viz` components have access to `onSetVariables`?), and why aren't we using `context` here like we do for `updateSource`?

packages/cms/src/components/cards/DimensionCard.jsx
```diff
@@ -78,6 +76,8 @@ export default class DimensionCard extends Component {
       onAlertCancel: () => this.setState({alertObj: false})
     };
 
+    // console.log(preview);
```
davelandry: Eww gross.

packages/cms/src/components/cards/VisualizationCard.jsx
```diff
@@ -102,6 +102,10 @@ class VisualizationCard extends Component {
     this.setState({minData, isOpen, alertObj, isDirty});
   }
 
+  onSetVariables(newVariables) {
+    if (this.props.onSetVariables) this.props.onSetVariables(newVariables);
```
davelandry: Same comments from `Viz.jsx` about the pass-through and not using context.

packages/cms/src/components/sections/Default.jsx
```diff
@@ -23,7 +23,7 @@ export default class Default extends Component {
         {/* caption */}
         <div className={`cp-default-section-figure${visualizations.length > 1 ? " cp-multicolumn-default-section-figure" : ""}`}>
           {visualizations.map((visualization, ii) =>
-            <Viz section={this} config={visualization} headingLevel={vizHeadingLevel} sectionTitle={title} slug={slug} key={ii} />
+            <Viz section={this} onSetVariables={onSetVariables} config={visualization} headingLevel={vizHeadingLevel} sectionTitle={title} slug={slug} key={ii} />
```
davelandry: Can this function be provided via `context`?

packages/cms/src/components/sections/Grouping.jsx
```diff
@@ -23,7 +23,7 @@ export default class Grouping extends Component {
         {/* caption */}
         <div className={`cp-grouping-section-figure${visualizations.length > 1 ? " cp-multicolumn-grouping-section-figure" : ""}`}>
           {visualizations.map((visualization, ii) =>
-            <Viz section={this} config={visualization} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={ii} />
+            <Viz section={this} onSetVariables={onSetVariables} config={visualization} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={ii} />
```
davelandry: Can this function be provided via `context`?

packages/cms/src/components/sections/Hero.jsx
```diff
@@ -1,5 +1,5 @@
 import React, {Component} from "react";
-import PropTypes from "prop-types";
+// import PropTypes from "prop-types";
```
davelandry: If PropTypes aren't used, please delete this line.

packages/cms/src/components/sections/InfoCard.jsx
```diff
@@ -40,7 +40,7 @@ export default class InfoCard extends Component {
           {visualizations && visualizations.length
             ? <div className="cp-section-content cp-info-card-section-viz">
               {visualizations.map((visualization, ii) => ii === 0
-                ? <Viz section={this} config={visualization} options={false} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={`${slug}-${ii}`} /> : ""
+                ? <Viz section={this} onSetVariables={onSetVariables} config={visualization} options={false} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={`${slug}-${ii}`} /> : ""
```
davelandry: Can this function be provided via `context`?

packages/cms/src/components/sections/MultiColumn.jsx
```diff
@@ -21,7 +21,7 @@ export default class MultiColumn extends Component {
           {paragraphs}
           {sources}
           {visualizations.map((visualization, ii) =>
-            <Viz section={this} config={visualization} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={`${slug}-${ii}`} />
+            <Viz section={this} onSetVariables={onSetVariables} config={visualization} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={`${slug}-${ii}`} />
```
davelandry: Can this function be provided via `context`?

packages/cms/src/components/sections/Section.jsx
```diff
@@ -205,6 +209,7 @@ class Section extends Component {
       paragraphs: layout === "Tabs" ? contents.descriptions : paragraphs,
       visualizations: !contents.sticky ? visualizations : [],
       vizHeadingLevel: `h${parseInt(headingLevel.replace("h", ""), 10) + 1}`,
+      onSetVariables: this.onSetVariables.bind(this),
```
davelandry: Can this function be provided via `context`?

packages/cms/src/components/sections/SingleColumn.jsx
```diff
@@ -23,7 +23,7 @@ export default class SingleColumn extends Component {
           {paragraphs}
           {sources}
           {visualizations.map((visualization, ii) =>
-            <Viz section={this} config={visualization} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={`${slug}-${ii}`} />
+            <Viz section={this} onSetVariables={onSetVariables} config={visualization} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title} key={`${slug}-${ii}`} />
```
davelandry: Can this function be provided via `context`?

packages/cms/src/components/sections/Tabs.jsx
```diff
@@ -98,7 +98,7 @@ export default class Tabs extends Component {
       </div>
 
       <div className="cp-tabs-section-figure">
-        <Viz section={this} config={visualization} key={tabIndex} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title}  />
+        <Viz section={this} onSetVariables={onSetVariables} config={visualization} key={tabIndex} slug={slug} headingLevel={vizHeadingLevel} sectionTitle={title}  />
```
davelandry: Can this function be provided via `context`?

packages/cms/src/profile/DimensionBuilder.jsx
```diff
@@ -17,8 +17,8 @@ export default class DimensionBuilder extends Component {
   componentDidMount() {
   }
 
-  onSelectPreview(slug, id) {
-    if (this.props.onSelectPreview) this.props.onSelectPreview(slug, id);
+  onSelectPreview(newPreview) {
+    if (this.props.onSelectPreview) this.props.onSelectPreview(newPreview);
```
davelandry: Similar to my comments on setVariables, do we need this `if` statement and can we use `context`?

packages/cms/src/profile/ProfileBuilder.jsx
```diff
@@ -666,6 +693,7 @@ class ProfileBuilder extends Component {
                 locale={locale}
                 localeDefault={localeDefault}
                 previews={previews}
+                onSetVariables={this.onSetVariables.bind(this)}
```
davelandry: Can this function be provided via `context`?

packages/cms/src/profile/SectionEditor.jsx
```diff
@@ -373,6 +377,7 @@ class SectionEditor extends Component {
                 previews={previews}
                 onDelete={this.onDelete.bind(this)}
                 type="section_visualization"
+                onSetVariables={this.onSetVariables.bind(this)}
```
davelandry: Can this function be provided via `context`?

packages/cms/app/helmet.js
```diff
@@ -1,6 +1,6 @@
 export default {
   link: [
-    {rel: "icon", href: "images/favicon.ico"}
+    {rel: "icon", href: "/images/favicon.ico"}
```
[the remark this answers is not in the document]
jhmullen: Not necessary as far as I know, so I'm removing it. The only reason I even changed it was actually creating a broken request to my custom `/image/` api endpoint, so would rather remove it anyway.

packages/cms/package.json
```diff
@@ -18,17 +19,22 @@
...
+    "canon-cms-migrate-0.7": "node scripts/migration/canon-cms-migrate-0.7.js"
   },
   "dependencies": {
     "@blueprintjs/datetime": "^3.7.1",
     "@blueprintjs/select": "^3.6.1",
+    "@datawheel/olap-client": "^1.0.15",
     "@datawheel/tesseract-client": "^0.1.1",
+    "@google-cloud/storage": "^3.1.0",
     "axios": "^0.18.0",
+    "base58": "^2.0.1",
```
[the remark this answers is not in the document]
jhmullen: This is used in MetaEditor.jsx for editing search members. As for why it's needed:

When you use the new "search for an image" feature in the CMS, the list that comes back from flickr has full-length (decimal) ids, not shortened ids (like the one in something like `https://flic.kr/p/m9bz7`). The short id is not returned any `getInfo` type methods - only the long id.

We need to store the shortened url in the db, and I wanted to keep the shortlink pattern of `https://flic.kr/p/m9bz7`.  To do so, I need to turn long decimal flickr ids into their shortened counterparts, which I discovered in flickr docs is a `Base58` conversion.

Does that answer your question?

packages/cms/package.json
```diff
@@ -39,14 +45,16 @@
     "react-clipboard.js": "^2.0.7",
     "react-quill": "^1.3.1",
     "react-sizeme": "^2.6.7",
-    "react-table": "^6.10.0"
+    "react-table": "^6.10.0",
+    "request": "^2.88.0",
```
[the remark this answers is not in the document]
jhmullen: The original code to server images in datausa used `res.sendFile`, but this only works for locally stored images (as they were for datausa). What was needed for the new image hosting was an image proxy, which would make a request to a url and do a simple passthrough of the results.

I finagled for a few hours with axios, trying to get the image to serve (I was trying to manually build the response using [this](https://github.com/axios/axios/issues/328) but had no luck. While researching I found [this](https://stackoverflow.com/questions/18432779/piping-remote-file-in-expressjs) and finally got the images serving. 

If you'd like, I can continue my research this week on doing the image serve pass-through with pure axios.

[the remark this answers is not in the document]
jhmullen: Never mind, I got it working with axios. Removed `request`.

packages/cms/scripts/migration/canon-cms-migrate-0.7.js
```diff
@@ -0,0 +1,97 @@
...
+        const {id, dimension, hierarchy, name, keywords} = oldrow;  
+        const newrow = newrows.find(d => d.id === id && d.dimension === dimension && d.hierarchy === hierarchy);
+        const {contentId} = newrow;
+        search_content.push({id: contentId, locale, name, keywords});
+      }
+      await dbnew.search_content.bulkCreate(search_content).catch(catcher);
+    }
+  }
+
+  console.log("Done.");
```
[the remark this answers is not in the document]
jhmullen: Added shell exit and better logging to all four scripts.

packages/cms/package.json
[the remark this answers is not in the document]
jhmullen: To expand on this, I just moved the Base58 logic from MetaEditor where it was to the serverside searchRoute where it really should be.  Let me know if you'd like me to make further changes.

packages/cms/src/api/imageRoute.js
```diff
@@ -11,9 +16,64 @@ const slugMap = {
...
+        return imageError();
+      }
+    }
+  });
+
+  /* LEGACY DATAUSA IMAGE ROUTE */
+
   const {parents} = app.settings.cache;
 
   app.get("/api/profile/:pslug/:pid/:size", (req, res) => {
```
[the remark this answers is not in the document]
jhmullen: Route removed, instructions added to readme (in this branch) and also the PR itself.

packages/cms/src/api/searchRoute.js
```diff
@@ -1,12 +1,151 @@
 const sequelize = require("sequelize");
+const yn = require("yn");
+
+const verbose = yn(process.env.CANON_CMS_LOGGING);
+const Flickr = require("flickr-sdk");
+const flickr = new Flickr(process.env.FLICKR_API_KEY);
```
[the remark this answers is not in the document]
jhmullen: Great catch, thank you. searchRoute.js now gracefully handles no flickrAPI key and gives Toast Warnings if users try to upload/use images.

packages/cms/src/api/searchRoute.js
```diff
@@ -1,12 +1,151 @@
...
+const {Storage} = require("@google-cloud/storage");
```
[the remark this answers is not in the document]
jhmullen: I put all these into an if statement behind `FLICKR_API_KEY`, and warn the user in a toast when they try to do images without configuring.

packages/cms/app/helmet.js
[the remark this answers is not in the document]
davelandry: This line shouldn't be removed, just the forward-slash that got added should be removed.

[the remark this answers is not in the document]
jhmullen: @davelandry When the forward slash isn't there, a request is made to `[currentpage]/images/favico.ico`.  Because of `routes.jsx`, the main page attempts to load a profile with a slug of `images` and an id of `favico.ico`, resulting in a failed profile lookup on every page load. Do you have a preference on how to handle this?

packages/cms/src/api/searchRoute.js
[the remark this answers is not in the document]
jhmullen: K, removed entirely

packages/cms/app/helmet.js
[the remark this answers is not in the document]
jhmullen: For now, I've added the line back in with no leading slash. Let me know your thoughts on the routing problem.

packages/cms/src/components/Viz/Options.jsx
jhmullen: I'm unsure of the history of this, but your reasoning is solid so I've pushed a reversion. @perpetualgrimace could you chime in on whether this change was made to cover some bug you discovered?

jhmullen:
Addressed all feedback, including the two comments above about the README and transparent.png

Issue #704, opened 2019-09-23
[vizbuilder] use "caption" if available, and then use "name"


davelandry:
This would apply to both:
- [ ] measures
- [ ] dimensions

jhmullen:
@frabarz is this still active?

PR #740, opened 2019-10-04
[vizbuilder/v0.2] updates labels to use caption property

frabarz:
related to #704


frabarz:
a192f1ee017c880de5a0cae620c33f629036c0b5 applies the fix for levels

PR #761, opened 2019-10-09
CDC fixes

### LINKED ISSUE #747: [cms] "Share.copy" and "Share.copied" appear in front-end data modal
These should be changed to the CMS namespace (like `"CMS.Options.Direct link"` a couple of lines down), and they will also need to be added to the default locale file at `src/i18n/canon.js`

### LINKED ISSUE #750: [cms] horizontal scrolling in view data table is broken in IE
from @gtdane:

> Regarding the VIEW DATA, the left to right scroll bar is not enabled until user scrolls to bottom of display

### LINKED ISSUE #752: [cms] sticky sections do not disappear in IE when entering a new grouping

perpetualgrimace:
A lil' bundle of bugfixes which _should_ close IE issues #750 and #752, as well as the non-IE issue #747.

Disclaimer: since the JS is currently not executing in IE, I'm unable to check my work in that browser. But I'm pretty confident they'll be fixed once we can get visualizations to print and calculate sticky section dimensions. 😬

- closes #750
- closes #752
- closes #747

### DIFF
--- a/packages/cms/src/components/Profile.jsx
+++ b/packages/cms/src/components/Profile.jsx
@@ -26,7 +26,8 @@ class Profile extends Component {
     this.state = {
       profile: props.profile,
       selectors: {},
-      loading: false
+      loading: false,
+      isIE: false
     };
   }
 
@@ -49,9 +50,9 @@ class Profile extends Component {
     };
   }
 
-  /** 
+  /**
    * Visualizations have the ability to "break out" and override a variable in the variables object.
-   * This requires a server round trip, because the user may have changed a variable that would affect 
+   * This requires a server round trip, because the user may have changed a variable that would affect
    * the "allowed" status of a given section.
    */
   onSetVariables(newVariables) {
@@ -91,7 +92,14 @@ class Profile extends Component {
   }
 
   render() {
-    const {profile, loading} = this.state;
+    const {isIE, profile, loading} = this.state;
+
+    // IE check needed for position: sticky fallback
+    if (typeof window !== "undefined") {
+      if (/*@cc_on!@*/false || !!document.documentMode) { // eslint-disable-line spaced-comment
+        this.setState({isIE: true});
+      }
+    }
 
     let {sections} = profile;
     // Find the first instance of a Hero section (excludes all following instances)
@@ -145,7 +153,10 @@ class Profile extends Component {
         {/* main content sections */}
         <main className="cp-main" id="main">
           {groupedSections.map((groupings, i) =>
-            <div className="cp-grouping" key={i}>
+            <div className="cp-grouping" key={i} style={isIE ? {
+              position: "relative",
+              zIndex: i + 1 // in IE, hide sticky sections behind the next grouping
+            } : null}>
               {groupings.map((innerGrouping, ii) => innerGrouping.length === 1
                 // ungrouped section
                 ? <Section
--- a/packages/cms/src/components/Profile.jsx
+++ b/packages/cms/src/components/Profile.jsx
@@ -10,6 +10,7 @@ import libs from "../utils/libs";
 import Hero from "./sections/Hero";
 import Section from "./sections/Section";
 import SectionGrouping from "./sections/components/SectionGrouping";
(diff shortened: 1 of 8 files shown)


packages/cms/src/components/Profile.jsx
```diff
@@ -92,7 +93,14 @@ class Profile extends Component {
...
   render() {
-    const {profile, loading} = this.state;
+    const {isIE, profile, loading} = this.state;
+
+    // IE check needed for position: sticky fallback
+    if (typeof window !== "undefined") {
+      if (/*@cc_on!@*/false || !!document.documentMode) { // eslint-disable-line spaced-comment
+        this.setState({isIE: true});
+      }
+    }
```
davelandry: using `setState` within the render function is a no-no... can this logic just be moved right into the constructor?

davelandry:
@perpetualgrimace also, if you format your text in the body of the PR with "closes #X" explicitly, it will automagically close those issues when this PR is closed 😄 

(so please go back and reformat the text)

perpetualgrimace:
Ah so it doesn't read that when I leave it in the commits, only in the PR?

packages/cms/src/components/Profile.jsx
perpetualgrimace: Ah okay. Because we'd be needlessly updating the state every time we need to re-render the component?

perpetualgrimace: https://github.com/facebook/react/issues/5342

perpetualgrimace: Apparently setState in the constructor is also a no-no. How about moving it to a `componentDidMount()` lifecycle method?

davelandry:
ahhh I see what you mean, yeah it should probably pick it up from the commits ¯\_(ツ)_/¯

perpetualgrimace:
ie-sticky-section

perpetualgrimace:
K, have verified both IE bugs are fixed.

packages/cms/src/components/Profile.jsx
```diff
@@ -27,10 +27,20 @@ class Profile extends Component {
...
 
+  // IE check needed for position: sticky fallback
+  componentDidMount() {
+    if (typeof window !== "undefined") {
+      if (/*@cc_on!@*/false || !!document.documentMode) { // eslint-disable-line spaced-comment
+        this.setState({isIE: true});
+      }
+    }
+  }
+
```
davelandry: Having this in the `componentDidMount` means that IE will render once without this, and then after the component mounts it will trigger this and re-render, causing an additional render cycle in IE. The React lifecycle is:
* `constructor`
* `render`
* `componentDidMount`

Moving this to the `constructor` would solve this (no need to initialize it as `false`), but I think a better solution would be to do this detection completely outside of the component, way up at the top of the file with the imports (ie. `const isIE` instead of using state)

packages/cms/src/utils/isIE.js
```diff
@@ -0,0 +1,16 @@
...
+    // will only return true for internet explorer
+    if (/*@cc_on!@*/false || !!document.documentMode) { // eslint-disable-line spaced-comment
+      isIE = true;
+    }
+  }
+
+  return isIE;
+}
+
+module.exports = isIE;
```
davelandry: Ahh, I see you didn't test this in non-IE browsers 😉 

This file exports a _Function_, which based on the way that you reference it in all of your components (`isIE`), it will always return `true` because a JavaScript _Function_ is a "truthy" variable. In it's current form, you would have to invoke this function to actually get the value (`isIE()`), but that would be running this code on every single render in every single usage. Ideally, this function code just be the following, but please test this in IE and Chrome to confirm:

```js
/** check if the current browser is internet explorer */
let isIE = false;

// make sure the browser window exists
if (typeof window !== "undefined") {
  // will only return true for internet explorer
  if (/*@cc_on!@*/false || !!document.documentMode) { // eslint-disable-line spaced-comment
    isIE = true;
  }
}

module.exports = isIE;
```

PR #763, opened 2019-10-09
[cms] remote search for Meta Panel

jhmullen:
The MetaEditor Panel can be quite slow on initial load, because I was loading ALL search members and then filtering with javascript.

This PR moves the querying to the server using limiting, which results in a much quicker experience for navigating large member tables.

### DIFF
--- a/packages/cms/src/member/MetaEditor.jsx
+++ b/packages/cms/src/member/MetaEditor.jsx
@@ -16,6 +16,7 @@ import Select from "../components/fields/Select";
 import "./MetaEditor.css";
 
 const IMAGES_PER_PAGE = 48;
+const ROWS_PER_PAGE = 10;
 
 class MetaEditor extends Component {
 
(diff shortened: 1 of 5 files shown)


packages/cms/src/member/MetaEditor.jsx
```diff
@@ -367,7 +370,7 @@ class MetaEditor extends Component {
    * Then call prepData to turn the sourceData into columns for the table.
    */
   hitDB() {
-    const searchGet = axios.get("/api/search/all");
+    const searchGet = axios.get("/api/search?q=&locale=all");
```
davelandry: does this mean that you can also completely delete the `"/api/search/all"` endpoint from `searchRoute.js`?

davelandry:
@jhmullen additionally, please reference @alexandersimoes's proposal in #762 and incorporate his changes (and my feedback in that PR) into this PR

jhmullen:
@davelandry search/all endpoint has been removed, pagesize option worked in, and I also added a "preview" column.

jhmullen:
@davelandry Empty Start & White bar fixed.

PR #766, opened 2019-10-11
[cms] 0.9 Modals

### LINKED ISSUE #706: [cms] allow visualizations to open a dialog box

jhmullen:
This is the CMS 0.9 Release, which adds support for Popover Modals. All the relevant documentation for how to use `openModal` (and the 0.8 feature `setVariables`) has been added to the cms README.

A new migration script has been added for upgrading from 0.8 to 0.9.

Closes #706


packages/cms/README.md
```diff
@@ -202,6 +203,58 @@ A Canon site often takes the form of DataCountry.io, and is made of **Profiles**
 
 ---
 
+## Advanced Commands
+
+For complex visualizations, it may be necessary to write custom code that causes another event to fire. There are two primary use cases here:
+
+### Opening another visualization
```
davelandry: this will more often be referred to as "interacting between visualizations" or "visualization click events changing page variables/state", we shouldn't lead with "opening another visualization" (that's not actually how they're using it in OEC)

davelandry: maybe name the section "Advanced Visualization Techniques", and then this sub-header "Modifying Page State"

jhmullen:
@davelandry Docs Updated.

PR #773, opened 2019-10-15
adds proper catch to cubeData population and supports empty case for …

jhmullen:
Two quick bug fixes:

The cubeData cache element can crash with an improper `CANON_CMS_CUBES` var. Adds a catch for that.

The MetaData table crashes if the user has no member data yet. Fixed.


davelandry:
@jhmullen looks great, feel free to do a release if you deem the hotfix is worthy of one!

PR #799, opened 2019-10-29
adds more robust payload extractor for simpleviz

### LINKED ISSUE #720: [cms] viz ui-mode doesn't work with APIs without "data" accessor
**motivation**
Last night I was a dumb dumb and thought I broke the cms, but it turns out I was just using an API that doesn't return an axios-style `.data` accessor. It'd be nice if we could use any old API to generate visualizations in the future.

(In the meantime, I will remember to use `/api/test` 😅)

jhmullen:
closes #720

### DIFF
--- a/packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
+++ b/packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
@@ -34,7 +34,7 @@ class SimpleVisualizationEditor extends Component {
     this.state = {
       object: {},
       rebuildAlertOpen: false,
-      payload: {}
+      payload: []
     };
   }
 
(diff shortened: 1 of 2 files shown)


packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -60,6 +60,18 @@ class SimpleVisualizationEditor extends Component {
...
+      return resp.data.data;
+    }
+    else if (resp.data) {
+      return resp.data;
+    }
+    else {
+      return resp;
+    }
+  }
+
```
davelandry: Since `resp` is always coming from axios, then we will ALWAYS need to do `resp.data` (because that's the actual payload). This if/else tree check for 3 things:
1. check for `resp.data` and `resp.headers` like d3plus-viz: https://github.com/d3plus/d3plus-viz/blob/master/src/data/load.js#L79
2. check for `resp.data`
3. return `resp`

jhmullen:
@davelandry Fixed logic per your specs

PR #800, opened 2019-10-29
auto-opens selectors and formatters on creation

### LINKED ISSUE #793: [cms] auto-open all editors on creation
This is currently hooked up to toolbox editors, but not page content.

jhmullen:
part of #793 

Also fixes bug with deleting selectors.

### DIFF
--- a/packages/cms/src/components/cards/SelectorCard.jsx
+++ b/packages/cms/src/components/cards/SelectorCard.jsx
@@ -35,6 +35,7 @@ class SelectorCard extends Component {
    */
   componentDidMount() {
     this.setState({minData: this.props.minData});
+    if (this.props.forceOpen) this.openEditor.bind(this)();
   }
 
   markAsDirty() {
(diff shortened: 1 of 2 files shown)


davelandry:
I tried to word the title of #793 to be more general so that it also includes things like new descriptions and new stats. How hard would it be to fold that into this PR?

jhmullen:
@davelandry This would require a bit more refactoring across the current state of 0.9, and most of that work would be nullified by the incoming `redux-refactor` branch. 

I would prefer to merge this stopgap of adding formatters and selectors, and then do more generic "open everything" work on the soon-to-be-done `redux-refactor` branch. 

Alternatively, I can do the "open everything" work on the current state of 0.9, and simultaneously port it to the differently behaving `redux-refactor` branch in parallel.  Let me know what is preferred.

davelandry:
Alright, we'll keep #793 open for now then.

PR #811, opened 2019-11-07
Issue 723 viz improvements

### LINKED ISSUE #723: [cms] rebuild UI viz mode
This will serve as a brain dump to start the conversation around a better UI mode for visualizations. I'm motivated to do the following things:

- [x] add "Plot" chart type, which would enable scatter plots (and double check that we are allowing every possible d3plus chart)
- [x] add access to more methods in UI mode (from a TBD list starting below)
  - [x] x/y axis titles
  - [x] x/y axis tick formatters
- [x] display the used data keys in the tooltip config by default
- [x] simplify variable selection by having "typeof" support

I've done a little thinking into how we should be structuring the like, sort of "JSON-config" for all of the available controls, and in my head it looks like this:

```js
const sharedMethods = [
  {
    key: "title",
    format: "Input", // full text input box saved as a string
    typeof: "number" // only show keys which have "number" values in the dropdown list
  }
];

const vizLookup: [
  {
    name: "Scatter Plot", // the name to be displayed in UI mode
    type: "Plot", // the actual d3plus component (the "type" key)
    methods: [
      {
        key: "groupBy",
        format: "Accessor", // save the result as `size: d => d[theUserSelectedKey]`,
        required: true,
        typeof: "id" // custom type defined for our project (see logic below)
      },
      {
        key: "x",
        format: "Accessor",
        required: true
      },
      {
        key: "y",
        format: "Accessor",
        required: true
      },
      {
        key: "size",
        format: "Accessor", // save the result as `size: d => d[theUserSelectedKey]`
        typeof: "number" // only show keys which have "number" values in the dropdown list
      }
    ]
  },
  ...
]
```

Here's my idea for the "id" typeof selector:
- hide all "non-dimension" keys from this list. we can make an assumption for tesseract/mondrian based on the "ID X" keys. 
- the dropdown should not show the names of the "ID X" keys, but just the "X" key names. The logic would then be: save the "ID X" key as the groupBy accessor, and save the "X" key as the label accessor
- the fallback is to show all keys

jhmullen:
This PR contains incremental improvements which add more options to the Simple Viz Builder, along with a more robust templating system.

As part of the PR review, I would appreciate some clicking around with sample data, if you have any "real" cms data lying around. I did my best with test coverage but a lot of functionality is changed here.

closes #723


packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -89,15 +73,33 @@ class SimpleVisualizationEditor extends Component {
   }
 
   compileCode() {
-    const {object} = this.state;
-    // If the user has put instances variables between brackets (e.g. <id> or <var>)
+    const {object, payload} = this.state;
+    const {type} = object;
+    const firstObj = payload && payload.data && payload.data[0] ? payload.data[0] : {};
+    const stripID = d => typeof d === "string" ? d.replace("ID ", "").replace(" ID", "") : d;
```
davelandry: I can't resist regex: `d.replace(/(ID\s|\sID)/g, "")`

packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -136,12 +174,52 @@ class SimpleVisualizationEditor extends Component {
...
+    else {
+      object.formatters[field] = e.target.value;
+    }
+    this.setState({object}, this.compileCode.bind(this));
+  }
+
+  getOptionList(method) {
+    const {payload} = this.state;
+    const firstObj = payload && payload.data && payload.data[0] ? payload.data[0] : {};
+    const isID = d => d.includes("ID ") || d.includes(" ID");
```
davelandry: `d.match(/(ID\s|\sID)/g, "")`

packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -136,12 +174,52 @@ class SimpleVisualizationEditor extends Component {
...
+    const allFields = Object.keys(firstObj);
+    const plainFields = Object.keys(firstObj).filter(d => !isID(d));
+    const idFields = Object.keys(firstObj).filter(d => isID(d));
```
davelandry: Instead of calculating `Object.keys(firstObj)` three times, just reference `allFields`.

packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -136,12 +174,52 @@ class SimpleVisualizationEditor extends Component {
...
+    let options = [];
+    if (!method.required) options.push({value: "manual-none", display: "none"});
```
davelandry: uppercase `display: "None"`

packages/cms/src/components/editors/vizLookup.js
```diff
@@ -0,0 +1,138 @@
...
+  commonMethods[key].tooltip = true;
+});
+const {groupBy, x, y, value, colorScale, label, subtitle, sum, xConfig, yConfig} = commonMethods;
+
+module.exports = [
+  {
+    name: "Area Plot", // the name to be displayed in UI mode
+    type: "AreaPlot", // the actual d3plus component (the "type" key)
+    methods: [groupBy, x, y, xConfig, yConfig, ...sharedMethods]
+  },
```
davelandry: This one's actually not really used (it's usually always `StackedArea`, so let's remove this for now.

packages/cms/src/components/editors/SimpleVisualizationEditor.jsx
```diff
@@ -207,8 +296,14 @@ class SimpleVisualizationEditor extends Component {
 
   render() {
     const {object, rebuildAlertOpen, payload} = this.state;
+    const {locale} = this.props;
+    const formatters = this.context.formatters[locale];
+    const formatterList = formatters ? Object.keys(formatters) : [];
```
davelandry: Please alpha-sort this list, it gets confusing when the list is rather large (I'm testing with CDC)

packages/cms/src/components/editors/vizLookup.js
```diff
@@ -0,0 +1,138 @@
...
+    required: false
+  };
+});
+
+// Promote certain keys to appear in the tooltip by default
+const tooltipKeys = ["x", "y", "value", "sum"];
+tooltipKeys.forEach(key => {
+  commonMethods[key].tooltip = true;
+});
+const {groupBy, x, y, value, colorScale, label, subtitle, sum, xConfig, yConfig} = commonMethods;
```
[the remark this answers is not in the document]
jhmullen: 1 & 2. Sounds good.

3. Title was originally in the now-removed `sharedMethods`. Title is kind of its "own thing" in that it is formatted to be outside of the normal keylisting:

image

Indeed, I'm excluding title from the schema config, because all I would do on the Front-end is just rip it out anyway and put it in its special non-looped place (and also filter the title OUT of the methods in every other place the keys get listed). I am of the opinion that global keys should, design-wise and schema-wise, stand alone visually in the SimpleVizConfig.  The vizLookup is for the "iterables" that are unique to any given viz.

4. They are called `xConfig.title` and `yConfig.title` right now, I use a template string.

jhmullen:
@davelandry Made requested changes and pulled in master.

packages/cms/src/components/editors/vizMethods.js
```diff
@@ -0,0 +1,93 @@
...
+  y: {
+    key: "y",
+    display: "y",
+    format: "Accessor",
+    required: true,
+    tooltip: true
+  },
+  yConfigTitle: {
+    key: "yConfig.title",
+    display: "y Axis Label",
```
davelandry: `display: "Y-Axis Label"`

packages/cms/src/components/editors/vizMethods.js
```diff
@@ -0,0 +1,93 @@
...
+  x: {
+    key: "x",
+    display: "x",
+    format: "Accessor",
+    required: true,
+    tooltip: true
+  },
+  xConfigTitle: {
+    key: "xConfig.title",
+    display: "x Axis Label",
```
davelandry: `display: "X-Axis Label"`

packages/cms/src/components/editors/vizMethods.js
```diff
@@ -0,0 +1,93 @@
...
+    key: "value",
+    display: "Value",
+    format: "Accessor",
+    required: true,
+    typeof: "number",
+    tooltip: true
+  },
+  x: {
+    key: "x",
+    display: "x",
```
davelandry: `display: "X-Axis"`

packages/cms/src/components/editors/vizMethods.js
```diff
@@ -0,0 +1,93 @@
...
+  },
+  xConfigTitle: {
+    key: "xConfig.title",
+    display: "x Axis Label",
+    format: "Input",
+    required: false
+  },
+  y: {
+    key: "y",
+    display: "y",
```
davelandry: `display: "Y-Axis"`

packages/cms/src/components/editors/vizMethods.js
```diff
@@ -0,0 +1,93 @@
...
+    key: "sum",
+    display: "Sum",
+    format: "Accessor",
+    required: true,
+    typeof: "number",
+    tooltip: true
+  },
+  topojson: {
+    key: "topojson",
+    display: "topojson",
```
davelandry: `display: "TopoJSON URL"`

jhmullen:
@davelandry Displays updated

PR #817, opened 2019-11-13
table viz pagination style

### LINKED ISSUE #783: [cms] Verify that pagination is visible when enabled in table viz, and style it
From Alex: "the payload i'm getting is ~900 items but the tables get cut at ~20 items. I added a config with showPagination:true but i don't see the pagination UI on the page"

perpetualgrimace:
Found out you can replace react-table's pagination controls with your own, so I replaced them with canon `Button`s (and did the same for the Meta table while I was at it). 

Tested and made sure the controls are in fact appearing when `showPagination` is set to `true` in viz config.

closes #783

### DIFF
--- a/packages/cms/src/components/Viz/Table.css
+++ b/packages/cms/src/components/Viz/Table.css
@@ -2,6 +2,10 @@
 
 .cp-table-viz-container .cp-viz-figure {
   height: auto !important;
+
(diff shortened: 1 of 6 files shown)


perpetualgrimace:
The dangers of testing only with small sample data. Good catch.

PR #831, opened 2019-11-19
makes search population blocking for #821

### LINKED ISSUE #821: id swaps on simple generator fetch requires refresh

jhmullen:
Closes #821 

@davelandry The reason for the above issue was that I was populating the search table asynchronously. Therefore, the Profile came back as created before any search rows were inserted, and no preview was being auto-selected.  This caused the preview to be empty, and `<id>` to not work in our first generator.

There are a few ways to think of how to address this:

1) You need to click your own preview AFTER adding a dimension. This is what people usually do, they search for a sample member in the search, and by the time they do this, the search has populated (partially, at least).

2) We could switch to the "I only ever fetch variables if you press a manual button" school of thought.  This would refuse to run if you hadn't picked a preview, which would prevent the behavior above.

3) I can block on profile creation, and wait for the members to populate, then auto-select one as normal.

I have chosen #3 for this PR. When a user adds a dimension, a Status bar asks that the user wait while the search members populate.  Once this is complete, the highest z-index is chosen automatically as usual. Though this can be a pain to wait, I can't pick a member until I'm done inserting search members, after all, that's the cause of the bug in the first place.

It's worth remembering that this interaction (adding dimensions) only happens a handful of times, early in CMS usage. 

If you want behavior #3, please review this PR. If not, let's chat more on how to handle this.

### DIFF
--- a/packages/cms/src/actions/profiles.js
+++ b/packages/cms/src/actions/profiles.js
@@ -45,6 +45,7 @@ export function swapEntity(type, id) {
 /** */
 export function modifyDimension(payload) { 
   return function(dispatch, getStore) {
+    dispatch({type: "SEARCH_LOADING"});
     const diffCounter = getStore().cms.status.diffCounter + 1;
     return axios.post("/api/cms/profile/upsertDimension", payload)
       .then(({data}) => {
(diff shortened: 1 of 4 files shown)


jhmullen:
@davelandry to follow up on this, I'm working an OEC issue and noticing this search operation can take minutes at a time, so this blocking may not be the best choice.

Do you have any thoughts on my comments above? Maybe I could even run the first 10 search insertions then return, so at least there's SOMETHING in there for the auto-select?

PR #854, opened 2019-12-03
adds duplicate catcher

### LINKED ISSUE #707: [cms] catch duplicate variables
If a user mistakenly exports a variable with the same name as a variable from a previous generator, it silently paves it. Provide some feedback to the user that this is occurring.

jhmullen:
Catches duplicate variables, Closes #707

I also backpacked a few bug fixes into this PR:

- fixed a bug that caused some variables not to show when switching to multi-language mode (due to a problem with how `diffCounter` was incremented)
- Simplified logic that GeneratorCards use to determine when to update their display panel
- Combined accidental duplicate cases in profile reducer, causing gen/mat deletions to misbehave.

### DIFF
--- a/packages/cms/src/components/cards/GeneratorCard.jsx
+++ b/packages/cms/src/components/cards/GeneratorCard.jsx
@@ -22,7 +22,8 @@ class GeneratorCard extends Component {
       displayData: null,
       secondaryDisplayData: null,
       alertObj: false,
-      isDirty: false
+      isDirty: false,
+      dupes: []
     };
   }
 
@@ -37,7 +38,6 @@ class GeneratorCard extends Component {
   componentDidUpdate(prevProps) {
     const {type, minData} = this.props;
     const {id} = minData;
-    const {localeDefault, localeSecondary} = this.props.status;
     // If the props we receive from redux have changed, then an update action has occured.
     if (JSON.stringify(prevProps.minData) !== JSON.stringify(this.props.minData)) {
       // If a gen/mat was saved, re-run fetchvariables for just this one gen/mat.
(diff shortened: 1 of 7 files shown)


packages/cms/src/components/cards/GeneratorCard.jsx
```diff
@@ -202,17 +208,20 @@ class GeneratorCard extends Component {
...
                   <LocaleName>{localeSecondary}</LocaleName>
-                  <VarTable dataset={secondaryDisplayData} />
+                  <VarTable dataset={secondaryDisplayData} dupes={dupes} />
                 </div>
               }
             </div>
           }
+          {dupes.length > 0 && 
+            <div style={{color: "red"}}>Warning: Variables reused!</div>
+          }
```
davelandry: @perpetualgrimace I would love to get your design review of this PR, maybe with Jimmy in person? 

Also, what do you think the text should say? I think we need to be more explicit here, something like "Variables exported by this generator conflict with another generator". Maybe it's just a little warning icon that you can see this message on hover?

packages/cms/src/components/variables/VarTable.jsx
```diff
@@ -16,9 +17,15 @@ export default class VarTable extends Component {
           <tbody className="cms-var-table-body">
             {Object.keys(dataset).map(k =>
               <tr className="cms-var-table-row" key={ k }>
-                <td className="cms-var-table-cell">
-                  { k }:
-                </td>
+                {dupes.includes(k) 
+                  ? <td style={{color: "red"}} className="cms-var-table-cell">
+                    { k }:
+                  </td>
```
davelandry: @perpetualgrimace need your review with this highlighting of a dupe variable

perpetualgrimace: @jhmullen

`<td className="cms-var-table-cell warning">`

`.cms-var-table-cell.warning { @mixin error-text; }`

packages/cms/src/components/cards/GeneratorCard.jsx
perpetualgrimace: @davelandry Jimmy Jam was attempting to add a tooltip but there was some funkiness with it repeatedly opening/closing itself. I think this will need some design time to come up with a non-clunky solution.

In the meantime, I agree that the message is a little vague, and an even better one might be something like `Warning: Variables ${ dupes.split() } conflict with another generator`

packages/cms/src/components/variables/VarTable.jsx
jhmullen: Fixed

packages/cms/src/components/cards/GeneratorCard.jsx
jhmullen: I fixed the mixin below, and changed the wording. It now looks like this:

image

I don't think there's need for `dupes.join()` because they are already highlighted in the VarTable.

@davelandry do we want to push this through as-is, or should we have @perpetualgrimace take the required design time for a better solution

davelandry: @jhmullen I'm okay with no tooltips and not using `dupes.join()`, but that highlighting is waaaaay too subtle. @perpetualgrimace what do you think?

perpetualgrimace: @jhmullen 

`<p className="cms-card-error u-font-xxs">{warningText}</p>`

`.cms-card-error { @mixin error-text; }`

PR #855, opened 2019-12-03
Vizbuilder 0.4

frabarz:
This PR merges the new version of vizbuilder into the core repo. This version enables interaction with tesseract-olap servers and mondrian servers at the same time, improves internal state and handling, usar accesibility and styling.


frabarz:
I finished the fix for the selector components, so now it should be ready for merging.

PR #862, opened 2019-12-05
[cms] Autocomplete

jhmullen:
I've finished what I'd consider to be my first pass at autocomplete functionality in the CMS.

This changes the text editor from Quill to Draft, and makes use of the `draft-js-mention-plugin` library.

You can use `@` to open a list of formatters, `{` to open a list of variables, and `[` to open a list of selectors.  Tabbing/arrow/enter all work for autocomplete.

There is some small weirdness with spaces, as the plugin I am using requires a leading space before the `@mention` kicks in (this is hardcoded into the library).  It's not quite "seamless" yet but I consider my testing to be miles ahead of closing the entire window and looking for what variable you made.

Please try it out in a testbed and let me know what you think.


perpetualgrimace:
Mostly I wanted to see it pop up in my dashboard, but I should do a design pass as well 😬

jhmullen:
@davelandry Unfortunately, some of the issues you've brought up are limitations of the mention library itself:

1) The library can support multi-character triggers, but it can't take Multiple different triggers. In my testing, you would need to choose either `{` or `{{` as the trigger.
2) I can't find a way to disable the trailing space, this appears to be baked into the library. However, for better or for worse, I have noticed that if the trailing space has no further sentence after it (as in a Stat, or if the user just forgot to remove it), the `toHTML` function calls it a `&nbsp;`.  This appears to be the only time the `&nbsp;` is put into the HTML, and it would be possible to strip this on write to avoid the mistake. Or, @perpetualgrimace would the eventual stripHTML that is called on the front-end, the one that removes `<p>` tags, handle this?
3) This is my biggest gripe and problem with the library as-is. There is a long [thread](https://github.com/draft-js-plugins/draft-js-plugins/issues/973) where people discuss how `mentionRegExp` has the space [HARD CODED](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/mentionSuggestionsStrategy.js#L40) before it. Changing this would appear to require a fork, but I'd like to maybe have you look it over with your regex powers at my desk? 💻 
4) Correct, if this transition were made, Quill would be removed (@perpetualgrimace not sure if that would require removing css cruft from anywhere)
5) 😬

davelandry:
@jhmullen for point number 4 about removing Quill: can you actually remove Quill from this PR? Both the wrapper component and the dependency?

davelandry:
And @jhmullen, given the limitations of Draft, for now let's use `{{` and `[[` as the triggers, not the single character.

jhmullen:
I've pushed some edits to the branch:

- `mentionTrigger` is set to double-brackets (`{{` or `[[`)
- QuillWrapper and `react-quill` have been removed
- I am manually stripping out the trailing `&nbsp;` cruft that happens when a mention is invoked at the end of a line (or when it is the only entity, as in a Stat)

@perpetualgrimace Could you please:

- Review RichTextEditor and DraftWrapper to confirm the embed looks good, as this was a pretty naive swap of quill and draft by me
- Add to RichTextEditor some kind of helper text, either permanently visible or behind a tooltip, that explains HOW to use these shortcuts - `{{` for variables, `[[` for selectors, and `@` for formatters.

perpetualgrimace:
✅design pass

PR #894, opened 2019-12-23
604 selectors as button groups

### LINKED ISSUE #604: [cms] display small selectors as button groups

perpetualgrimace:
I think I implemented this correctly, but please test before releasing.

closes #604

### DIFF
--- a/packages/cms/src/components/ProfileRenderer.jsx
+++ b/packages/cms/src/components/ProfileRenderer.jsx
@@ -110,6 +110,8 @@ class ProfileRenderer extends Component {
   }
 
   onSelector(name, value) {
+
+    console.log(name, value);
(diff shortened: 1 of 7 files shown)


davelandry:
@jhmullen can you take care of reviewing and merging this PR?

jhmullen:
@davelandry I tested this and the behaviors are correct, but I need to pass it back to you for two things:

- I put in a "label" field into ButtonGroup, trying to copy the one from Selector.css. Can you make sure I did this right (see line 10 of ButtonGroup.css)
- I can't get the button to STAY selected when I click off it - either on the page, or on another selector.  `active` is still true... but I imagine some other tabIndex thing may be overriding?

Thanks

PR #911, opened 2020-01-14
Issue 905 searchfix

### LINKED ISSUE #905: Rewrite Search Population
The populateSearch endpoint needs improvements:

- rewrite levels/hierarchy crawler to take the set union of all hierarchies
- change to bulkCreate/bulkUpdate from slow await forloop

jhmullen:
Does not require a major version release.

Two primary changes here:

- As discovered via the `Metro Areas` bug in DM, we now crawl through *ALL* hierarchies and take the set union of the discovered levels before hitting the cube.
- Search ingest changed from a looping `await` to a bulk upsert, speeding up ingest by an order of magnitude. Due to limitations in the search table and in sequelize (namely, search table uses a composite key) we were unable to use the `bulkCreate` method, and instead generate and execute a raw SQL query with an `ON CONFLICT` clause. This is not ideal, but the speed gains are undeniable. 

**NOTE:** I'd request that users take a backup of their cms db before running the new ingests, and if possible, contact me directly to supervise new ingests so I can confirm functionality. I've done as much testing as I can but would like to make sure it behaves nicely in the wild.

Closes #905

### DIFF
--- a/packages/cms/src/api/cmsRoute.js
+++ b/packages/cms/src/api/cmsRoute.js
@@ -55,7 +55,7 @@ const isEnabled = (req, res, next) => {
 
 const catcher = e => {
   if (verbose) {
-    console.error("Error in cmsRoute: ", e);
+    console.error("Error in cmsRoute: ", e.message);
   }
   return [];
 };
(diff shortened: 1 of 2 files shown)


davelandry:
Great work here @jhmullen. I leave it to you to release and notify the Datawheel users in Slack of your **NOTE** from above 👆

PR #928, opened 2020-01-28
[cms] fixes bug with image hosting wires crossed

jhmullen:
Victor discovered a bug where images with the same member id were not showing different images, only the "first" one. This was because I hadn't worked the new `cubeName` uniqueness into the image hosting logic. This PR fixes that bug.


jhmullen:
@davelandry Nothing need be changed about the README - it's simply that in my code at the route level, I was naively selecting a member by dimension ONLY for my member lookup. The same slug/id combo that one always used on the front-end is still valid - I just look it up correctly now.

There is a secondary type of lookup used in the Meta Editor, one that uses `dimension`, `cubeName` and `id` to find a member, which does not require directly knowing to which profile this member belongs. This is used ONLY in the meta editor (not front-end) and has already been updated to make use of cubeName.

PR #941, opened 2020-02-11
[cms] Automatic Parents Lookup

### LINKED ISSUE #910: [cms] parents for top-line generator and URLswap

jhmullen:
Closes #910 

Adds a `parents` object to the Attributes generator, allowing for subsequent generator API calls to make use of its contents.

Adds functionality and documentation for string-based object crawling in Generator APIs, e.g., `/api?hierarchy=<parents[0].value>`

Also adds locale-specific "name" to the Attributes generator.

Notes:

- The parents call currently takes up to 10 seconds. Because having the parents object is necessary for subsequent generators, parent fetching blocks all generators from running until this operation is complete. @MarcioPorto is looking into this at the cube level, but @alexandersimoes should be aware that upgrading OEC to a version of the CMS that has this PR in it will add to load times (until the parents endpoint is examined)
- @alexandersimoes You will also want to examine conflicting variables after upgrading. If you had specific generators for finding the `name` or the `parents` in the past, you will need to clean those up as the Attributes Generator should handle them now.

### DIFF
--- a/packages/cms/README.md
+++ b/packages/cms/README.md
@@ -10,6 +10,7 @@ Content Management System for Canon sites.
 * [Environment Variables](#environment-variables)
 * [Sections](#sections)
 * [Search](#search)
+* [Advanced Generator Techniques](#advanced-generator-techniques)
 * [Advanced Visualization Techniques](#advanced-visualization-techniques)
 * [Authentication](#authentication)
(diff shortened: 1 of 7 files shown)


alexandersimoes:
Great notes, will keep them in mind when upgrading @jhmullen.

PR #958, opened 2020-02-24
866 related profiles

### LINKED ISSUE #866: [cms] follow-up profile footer

jhmullen:
Closes #866 

Adds a neighbors array to the profile payload, which can be passed to Related.jsx to show recommended neighboring profiles.


packages/cms/src/components/sections/Related.jsx
```diff
@@ -0,0 +1,62 @@
...
+        title: profile.length === 1 ? profile[0].name : `${profile[0].name} / ${profile[1].name}`,
+        link: linkify(router, profile, locale),
+        images: profile.map(d => ({src: `/api/image?slug=${d.slug}&id=${d.id}&size=thumb`}))
+      }));
+
+      this.setState({tiles});
+    }
+
+    return null;
+  }
```
davelandry: @jhmullen a word of the wise: because nothing renders on initial mount, and you/James is formatting the data in componentDidMount, that means that these tiles aren't rendered in the server-side payload... which defeats the whole SEO purpose of these links! I'm refactoring this now to use the Tile component that I made for the search, so no worries about fixing this (I'll take care of it)

PR #961, opened 2020-03-02
adds slug injector and documentation

### LINKED ISSUE #957: [cms] fold slugs into user-specified generator logiclayer requests

jhmullen:
closes #957

### DIFF
--- a/packages/cms/README.md
+++ b/packages/cms/README.md
@@ -485,8 +485,6 @@ If the pieces are the same, one parameter may be used:
 
 `&slugs=Product`
 
-**Warning** This slug parameter **will not work** if your result set has more than 10 elements. Remember to use `limit` to keep your generator calls tiny!
-
 ---
 
 ## Advanced Visualization Techniques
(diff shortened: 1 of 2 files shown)


packages/cms/README.md
```diff
@@ -466,6 +466,27 @@ However, be warned that this is not "true" javascript, merely string manipulatio
...
+
+These parameters should be added to the generator API, using colons to separate the two required pieces:
+
+`&slugs=Exporter:Country,HS Product:HS4`
+
+If the pieces are the same, one parameter may be used:
+
+`&slugs=Product`
+
+**Warning** This slug parameter **will not work** if your result set has more than 10 elements. Remember to use `limit` to keep your generator calls tiny!
```
davelandry: Hmm, I understand why we would want to put in a limit, but 10 seems kind of... small. @jhmullen did you do any testing to see how much of a performance hit it is to do it on say, 100 rows? My gut says:
- this default number could/should be a little higher
- there should be an env var that allows users to change this limit

jhmullen: @davelandry I'd rather take off the limit entirely than add a config. There is only one `db.search` call per colon-separated pair (I use findAll and make a hashmap). In my OEC testing, performance does not seem to degrade. I think it's best to remove the limit entirely, this optimization is more liable to cause confusion than help us in the speed department.

PR #962, opened 2020-03-05
Issue 952 titlesearch

### LINKED ISSUE #952: [cms] click splash title to open faceted search
The idea here is that when a user is on a profile page, they may want to change to a different profile of the same "type" (dimension). The most natural way to do this from a user perspective is to click the title of the page they are on and open a search faceted on the specific dimension they are viewing (similar to the compare search in Data USA). We will also need to keep bilateral profiles in mind here which could have 2 different dimension (or the same) like country x product and country x country.

jhmullen:
Closes #952 

Overloads the ProfileSearch component to receive a formatting callback, so it may be used for single-dimension searches.

@davelandry Can you please test and design-pass this before merging?

### DIFF
--- a/packages/cms/src/components/fields/ProfileSearch.jsx
+++ b/packages/cms/src/components/fields/ProfileSearch.jsx
@@ -227,7 +227,7 @@ class ProfileSearch extends Component {
 
     const {router} = this.context;
     const {active, loading, query, results} = this.state;
-    const {locale} = this.props;
+    const {locale, mode, metaOptions, resultFormat} = this.props;
     const {
       availableProfiles,
       display,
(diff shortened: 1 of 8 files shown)


packages/cms/src/components/fields/ProfileSearch.jsx
```diff
@@ -238,6 +238,9 @@ class ProfileSearch extends Component {
       profileTitles
     } = this.props;
 
+    let dimensionResults = [];
+    if (mode === "dimension" && formatResults && results && results.results) dimensionResults = formatResults(results.results);
+
```
davelandry: How would you feel about moving this usage of `formatResults` directly into the axios get? We could even make the default value be `resp => resp` so that we could just run it on the payload regardless.

packages/cms/src/components/fields/ProfileSearch.jsx
```diff
@@ -279,38 +282,52 @@ class ProfileSearch extends Component {
...
+                      <li key={`r-${j}`} className="cms-profilesearch-list-item">
+                        <Link to={linkify(router, result, locale)} className="cms-profilesearch-list-item-link">
+                          {result.map(d => formatTitle(d.name)).join(` ${joiner} `)}
+                          <div className="cms-profilesearch-list-item-sub u-font-xs">{profileTitles[result.map(d => d.slug).join("/")] || formatCategory([result])}</div>
+                        </Link>
+                      </li>
+                    )}
+                  </ul>;
+                }
+                else {
```
davelandry: ehhhh I'm not a fan of duping this "list" formatting code. How would you feel about ditching the `mode` prop all together, and just adjusting the way you `formatResults` to include like, a dummy column?

packages/cms/src/components/fields/ProfileSearch.jsx
jhmullen: Done

jhmullen:
@davelandry 

Updated to both use a `resp => resp` formatter function and to only use one list renderer for both profile and dimension modes.

I have also backpacked the `jwt` authentication fixes into this PR - addressing @alexandersimoes 's problem with populateSearch and heading off other future problems with the `:parents` endpoint

jhmullen:
@davelandry 

Removed `mode` entirely as requested. The external `formatResults` how handles transforming the data (and filling in the fixed column) so it is identical to the `grouped` list that comes over the wire - allowing the unified list renderer to handle both cases.

PR #971, opened 2020-03-17
fixes sorting and normalizes on the fly

### LINKED ISSUE #968: [cms] profileSearch not returning variants in default empty mode

jhmullen:
closes #968 

Fixes entity sorting in the "nothing typed yet" autofill of the search panel.

Also required me to fix zvalues - we now normalize them on the fly (in this empty-search case) to avoid needing a new db column to store them.

### DIFF
--- a/packages/cms/src/api/searchRoute.js
+++ b/packages/cms/src/api/searchRoute.js
@@ -269,17 +269,10 @@ module.exports = function(app) {
           order: [["zvalue", "DESC NULLS LAST"]],
           limit
         });
-        if (!results.results[dc.dimension]) results.results[dc.dimension] = [];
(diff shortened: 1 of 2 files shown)


jhmullen:
@davelandry 

Ideally, normalizing on import is a better solution. However because normalization is not idempotent, I was concerned about edge cases and "separate profile with non-overlapping levels" re-ingests accidentally re-normalizing the values. To keep testing/deployment time at a minimum I opted for on-the-fly.

However, I may be playing it too safe here - I'll update this branch with a normalize-on-import.

jhmullen:
@davelandry 

I've changed the normalization logic as requested to be permanent, on import. Note: After this is merged, I'll need to go in with marcio and update all the existing zvalues in-place in the OEC.

PR #985, opened 2020-04-13
[cms] adds alert before leaving page

### LINKED ISSUE #969: [cms] alert modal when leaving the page

jhmullen:
closes #969

### DIFF
--- a/packages/cms/src/components/interface/Header.jsx
+++ b/packages/cms/src/components/interface/Header.jsx
@@ -98,14 +98,6 @@ class Header extends Component {
     this.setState({profileTarget: e.target.value});
   }
 
-  // When the user leaves the page to view the front-end profile, clear the back/reload blocker that was 
-  // only meant for the CMS (prevents the popup from erroneously occuring when the user tries to leave the front-end)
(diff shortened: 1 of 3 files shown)


packages/cms/src/components/interface/Header.jsx
```diff
@@ -98,6 +98,14 @@ class Header extends Component {
...
   }
 
+  // When the user leaves the page to view the front-end profile, clear the back/reload blocker that was 
+  // only meant for the CMS (prevents the popup from erroneously occuring when the user tries to leave the front-end)
+  clearBlocker() {
+    if (typeof window !== "undefined") {
+      window.onbeforeunload = () => undefined;
+    }
+  }
+
```
davelandry, line 108: wouldn't it be better to add an `componentWillUnmount` method that removes the `onbeforeunload` method? There may be some CMS implementations that have full-site nav above the CMS builder, and we can't add events to those links.

jhmullen: This is a great suggestion, I forgot about that part of the lifecycle. Moved, pushed, and ready for review.

PR #1023, opened 2020-06-04
Prepare conversion to monorepo

frabarz:


frabarz:
### `example` folder
The main benefit of a monorepo is the ability to link all packages live during the development, so all changes can be viewed/tested quickly. The main reason for me to moving the dev app for canon-core was to separate the files intended for core from the ones intended for demo (as I was very confused at start on which was which), but the other reason has to do with this ability. This example environment let us test all canon plugins with one another an see if there's synergy issues.

### @datawheel/eslint-config
I published this package a few months ago as a concept, but has worked nicely. We publish this package and use company-wide on our projects. Then we don't have duplicate configurations on every single repo, be it canon or a project scaffolded by canon. All configurations are kept in sync too, so if we need to update it, is centralized.

### docs split up
This was an idea, looking forward to make a static documentation site. I agree is harder to find something on this structure, but in a generated website we could use a search function. The main issue for me is that the [this page](https://www.npmjs.com/package/@datawheel/canon-core) is too long and discourages the use by other people, as it looks there's a lot to configure.

### jsconfig.json
`.vscode/settings.json` and `jsconfig.json` have different roles.  
The former is a local version of the IDE global settings, which currently has settings complementary to the eslint format we use. We can setup snippets and other canon-specific helpers that work in the IDE.
The latter is a javascript compilation config file, a cousin of the typescript compilation config file `tsconfig.json`, but it doesn't have to do with typescript directly. This file is intended to tell the typescript compiler (which VSCode uses as source of Intellisense) that all the sibling files in this folder belong to a same project, so it lets VSCode do automatic refactors when a file is moved or renamed. It also tells it where to look for files, how to resolve the $app/$root aliases, and to complain about javascript errors with the squiggly red underline.

### extras
The `.gitignore` in a scaffolded project and in the root of the monorepo is not exactly the same, but the changes I did were done before I move the example folder so I might missed something.
Eslint resolves its configuration upwards until it finds a config setting, which is on the package.json file at the root of the monorepo. Basically, all packages + example share the same [eslint-config](https://www.npmjs.com/package/@datawheel/eslint-config).

frabarz:
Closing without merge as discussed, will split the contents in smaller PRs.

PR #1027, opened 2020-06-08
moves core source to own package

frabarz:
Also:
* adds webpack aliases for $root and $app
* updates webpack config to resolve in a monorepo
* updates demo app references
* updates release script


packages/core/package.json
```diff
@@ -0,0 +1,148 @@
...
+    "webpack-bundle-analyzer": "^3.4.1",
+    "webpack-cli": "^3.3.7",
+    "webpack-dev-middleware": "^3.7.0",
+    "webpack-hot-middleware": "^2.25.0",
+    "webpackbar": "^4.0.0",
+    "yml-loader": "^2.1.0",
+    "yn": "^3.0.0"
+  },
+  "optionalDependencies": {
+    "d3plus-react": "^0.5.9"
```
davelandry, line 146: this got added to `optionalDependencies`, but is still listed in `dependencies`?

packages/core/webpack/config/loaders.js
```diff
@@ -31,47 +31,49 @@ module.exports = props => {
...
+    require.resolve("@babel/plugin-transform-react-constant-elements")
 
   ];
 
   if (process.env.NODE_ENV === "development") {
-    babelPlugins.push("react-hot-loader/babel");
-    babelPlugins.push("@babel/plugin-transform-react-inline-elements");
+    babelPlugins.push(
+      require.resolve("react-hot-loader/babel"),
+      require.resolve("@babel/plugin-transform-react-inline-elements")
```
davelandry, line 75: I'm totally fine with adding in these `require.resolve` wraps, but just out of curiousity, what is this protecting against?

packages/core/webpack/dev-client.js
```diff
@@ -14,6 +14,7 @@ const commonLoaders = require(loaderPath);
 
 process.traceDeprecation = true;
 
+/** @type {import("webpack").Configuration} */
```
davelandry, line 17: Is this some sort of JSDoc import of the specific `type` we're exporting here? Cool!

packages/core/jsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "jsx": "preserve",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "strictNullChecks": true,
+    "target": "es2017",
+  },
+  "exclude": [
+    "node_modules"
+  ]
+}
```
davelandry, line 14: iIs `jsconfig.json` the new preferred method over `.vscode/settings.json`? Is that what this file is? I don't mind moving to this, but do we need both?

packages/core/package.json
frabarz: oh, dang
I might missed that on the combination.

packages/core/webpack/config/loaders.js
frabarz: Is because of the monorepo
Package managers install the dependencies in their own package/folders/node_modules, but then the crosslinking makes the resolution relative to the node_modules in the current working directory. This part is not subject to the webpack.resolve configuration, so this is the most reliable way I found that doesn't interfere when installing as a package.

packages/core/webpack/dev-client.js
frabarz: It's a JSDoc typing. I've been using these for months, Intellisense caughts them and tell me about allowed types and errors (if there's a jsconfig.json file in the project).

packages/core/jsconfig.json
frabarz: `.vscode/settings.json` and `jsconfig.json` have different roles.
The former is a local version of the IDE global settings, which currently has settings complementary to the eslint format we use. We can setup snippets and other canon-specific helpers that work in the IDE. This file however can only be read by VSCode.
The latter is a javascript compilation config file, a cousin of the typescript compilation config file tsconfig.json, but it doesn't have to do with typescript directly. This file is intended to tell the typescript compiler (which VSCode uses as source of Intellisense) that all the sibling files in this folder belong to a same project, so it lets VSCode do automatic refactors when a file is moved or renamed. It also tells it where to look for files, how to resolve the $app/$root aliases, and to complain about javascript errors with the squiggly red underline. This file can be interpreted by any IDE that uses intellisense as helper (I think Atom and a bunch more can do it)

davelandry: Wow this was an amazing explanation. Thank you! Can you add this file into the scaffold directory as well so all future projects will get it?

PR #1032, opened 2020-06-11
Moves scaffolding to own package

### LINKED ISSUE #718: move redux-logger middleware from core to canon.js scaffolding file
During construction, [redux-logger accepts parameters](https://www.npmjs.com/package/redux-logger#recipes) to customize the middleware behavior. Since it's part of canon's internals, there's no way to set them unless you try to get creative with canon.js reduxMiddleware function.

I propose moving the whole initialization of redux-logger to the canon.js file of each project, through the canon.js scaffolding file. This way it serves as an example of how to add middlewares to any canon project, gives full control on enabling/disabling/passing parameters to it, and simplifies canon-core code.

The modification would be move line 26 of [`src/storeConfig.js`](https://github.com/Datawheel/canon/blob/master/src/storeConfig.js#L26) to the [`bin/scaffold/canon.js`](https://github.com/Datawheel/canon/blob/master/bin/scaffold/canon.js) file, and removing the references to the `__LOGREDUX__` environment variable.

`canon.js`
```js
const {createLogger} = require("redux-logger");

module.exports = {
  reduxMiddleware(applyMiddleware, middleware) {
    if (__DEV__ && !__SERVER__) {
      middleware.push(createLogger({
        collapsed: (getState, action, logEntry) => !logEntry.error
      }));
    }
    return applyMiddleware(...middleware);
  }
};
```

frabarz:
This package moves the scaffolding files to their own packages for ease the process of creating a new canon app.
Additionally, it adds the redux-logger middleware to the canon.js file as a demonstration, updates the vscode settings file, adds a special jsconfig.json file, and replaces the .eslintrc file in the scaffolding with a reference to the @datawheel/eslint-config package.
Closes #718


packages/create-canon/scaffold/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "canon-app",
+  "description": "A new instance of a canon app.",
+  "version": "1.0.0",
+  "private": true,
+  "main": "./index.js",
+  "scripts": {
+    "dev": "canon-dev",
+    "build": "canon-build",
```
davelandry, line 9: We found most sites need to also increase the `max-old-space-size` of node before building, do you have a strong preference on removing that from the default build script? Here's the old version:
`"build": NODE_OPTIONS=--max-old-space-size=4096 canon-build`

packages/create-canon/scaffold/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "locale": "canon-locale"
```
davelandry, line 10: I know it's not specifically a canon script, but don't you think it's worth having the start script in here as well?

`"start": "node index.js"`

packages/create-canon/scaffold/app/pages/Home.jsx
```diff
@@ -12,3 +12,5 @@ export default class Home extends Component {
   }
 
 }
+
+export default Home
```
davelandry, line 16: does this pass the eslint? missing an ending semicolon? 😏

packages/create-canon/README.md
```diff
@@ -0,0 +1,15 @@
+# @datawheel/create-canon
+
+A quick script to create a new canon app.
+
+## Usage
+
+```bash
+npm init @datawheel/canon
```
davelandry, line 8: I'm just learning about this `init` command, wouldn't this be `npm init canon` if our package is called `create-canon`?

frabarz: basically, yes
`npm init something` will look for a package called `create-something` on the registry

packages/create-canon/scaffold/app/pages/Home.jsx
frabarz: :pikachu:

packages/create-canon/scaffold/package.json
frabarz: ah, yeah, definitely goes here

packages/create-canon/scaffold/package.json
frabarz: The problem with that command is that only works on POSIX, but neither CMD nor PowerShell accept inline environment variables on commands. We would need to use the `cross-env` package to set it.
But in any case, I'd look into what causes the building process to use so much memory instead of this, because there might be a case where the user tries to run it in a computer without that much memory anyway.

PR #1058, opened 2020-07-09
[cms] Feature - Allowed Generators and Hidden Profiles (0.14)

### LINKED ISSUE #1074: [cms] Hidden Profiles
Add a visible true/false to profiles for pre-release work.

jhmullen:
... I couldn't help myself last night :trollface: 

🔥  WARNING: This requires a 0.14 database migration.  🔥 

Adds an allowed drop-down to generators and materializers. Generators can use magic generator variables (or customAttributes) for allowed, Materializers can use any generator variables.

`allowed` will determine whether the generator runs or not for this particular member selection.

I also added the ability of all cards to show visually when they are not allowed for a given member:

image

@davelandry you may want to do a visual pass on this - the relevant css is the `is-not-allowed` class in Card.css.

This also closes #1074 by allowing hiding of entire profiles


jhmullen:
For tracking and posterity, I am linking this to issue #1079

jhmullen:
OK @davelandry I think we're all set here. 

There are four new database columns here, gen/mat allowed and profile/variant visibility, requiring a migration script. As the model changes are fairly minor, I would prefer to make the change in-place on the dbs for each flagship product (part of my upcoming sqllite research will include abandoning these migration scripts entirely). I can work directly with each product lead to handle this.

Don't forget to combine this PR with `feature-1070-pdf`, and don't forget my request for a little bit of visual help (described in initial PR).

PR #1060, opened 2020-07-09
[core] Web workers in fetchData.

### LINKED ISSUE #1009: [core] use webworkers in default fetchData function
I was thinking this could be a good place to test the new web workers idea. Can we move the `axios.get` and the throttling into a web worker? https://github.com/Datawheel/canon/blob/master/src/actions/fetchData.js#L35

palamago:
* Web workers in fetchData. Close #1009 
* New helper function to create and run custom needs in a web worker.


packages/core/src/actions/fetchData.js
```diff
@@ -59,4 +82,29 @@ function fetchData(key, url, format = d => d, config = {}, useCache = true) {
...
+    try {
+      launchWorkerWrapper(fn, callback);
+    }
+    catch (error) {
+      console.error(error);
+      reject(error);
+    }
+  });
+}
+
```
davelandry, line 110: Is this function essentially just a `need` without any axios get? I don't necessarily see the use for this... but I DO see the need for a helper function that simply runs the supplied function and returns a promise (basically this, but without all the `key` and store stuff for needs). What do you think about:
1. simplifying this to just run the supplied function arguement in a web worker
2. moving it to a separate file within `src/helpers`

packages/core/webpack/config/loaders.js
```diff
@@ -72,7 +72,7 @@ module.exports = props => {
 
   if (process.env.NODE_ENV === "development") {
     babelPlugins.push(
-      require.resolve("react-hot-loader/babel"),
+      // require.resolve("react-hot-loader/babel"),
```
davelandry, line 75: does the new worker-loader break hot reloading? why is this commented?

packages/core/webpack/config/loaders.js
```diff
@@ -127,6 +127,11 @@ module.exports = props => {
     {
       test: /\.(scss|sass|css)$/i,
       use: !props.extract ? ["iso-morphic-style-loader"].concat(cssLoaders) : [MiniCssExtractPlugin.loader].concat(cssLoaders)
+    },
+    {
+      test: /\.worker\.js$/,
+      loader: "worker-loader",
+      options: {inline: true, fallback: false, publicPath: "/src/actions/"}
```
davelandry, line 134: have you tested a prod build? I see this `publicPath` being relative and get paranoid 😅

packages/core/webpack/dev-client.js
```diff
@@ -11,15 +11,16 @@ const appPath = path.join(appDir, "app");
...
 delete require.cache[loaderPath];
 const commonLoaders = require(loaderPath);
+const context = path.join(__dirname, "../src");
 
 process.traceDeprecation = true;
 
 /** @type {import("webpack").Configuration} */
 module.exports = {
-  devtool: "eval",
+  devtool: "source-map",
```
davelandry, line 20: ooooo what's this do? jeje

packages/core/webpack/dev-client.js
```diff
@@ -31,7 +32,8 @@ module.exports = {
   output: {
     path: assetsPath,
     filename: "[name].js",
-    publicPath
+    publicPath,
+    globalObject: "this"
```
davelandry, line 36: also curious about what `this` does 😏

packages/core/src/actions/fetchData.js
```diff
@@ -1,56 +1,79 @@
 import axios from "axios";
 import PromiseThrottle from "promise-throttle";
 import {encodeChars} from "../helpers/fixChars.js";
+import "@babel/polyfill";
```
davelandry, line 4: `"@babel/polyfill"` is actually included as an entry point in the prod and dev client webpack configs... maybe it just needs to also be added to the server webpack configs instead of this import? I just get nervous importing this in two places (potential code duplication, etc)

packages/core/package.json
```diff
@@ -142,5 +143,8 @@
     "webpackbar": "^4.0.0",
     "yml-loader": "^2.1.0",
     "yn": "^3.0.0"
+  },
+  "devDependencies": {
+    "worker-loader": "^2.0.0"
```
davelandry, line 148: this should be moved to `dependencies` with the other loaders (`devDependencies` are not installed when you install canon-core in a different repo)

palamago: You are right.

packages/core/src/actions/fetchData.js
palamago: Good one. I added to webpack, in server build. Looks better.

packages/core/webpack/config/loaders.js
palamago: Yes, I did. Haha, same here.

packages/core/webpack/dev-client.js
palamago: Oh! it was a test. 
This helps us to build the blueprint components correctly with their proper `.map` file. I was tired of all of those annoying  warning messages. -just in dev mode btw-. It is not webworker related. I'll double check if it worth it.

packages/core/webpack/config/loaders.js
palamago: Yeah, it breaks the hot-reloading. :(
 I'm on it.

davelandry:
@jhmullen ...should we need be reviving this? or have we deviated too much since here?

palamago:
We discussed with Jimmy a bit about it: I think we are too far from master now. So, delete the PR but keep the branch alive was my suggestion. Hopefully, with latest webpack we can create the production build properly now and support the workers.

PR #1068, opened 2020-07-17
[cms] download attachments

rhauckdot:
Added functionality to Options.jsx to accept dataAttachments prop from Viz config as either a string or array. Passed value is included into the JSZip call.


packages/cms/src/components/Viz/Options.jsx
```diff
@@ -115,9 +115,66 @@ class Options extends Component {
...
+              const dAName = dataAttachments[i].split("/").pop();
+              zip.file(`${dAName}`, blob, {binary: true});
+            }
+          }
+          zip.generateAsync({type: "blob"})
+            .then(content => saveAs(content, `${filename(title)}.zip`));
+        });
+      }
+
+    }
```
davelandry, line 172: @rhauckdot this can be written simpler, by coercing `dataAttachments` into an Array (even if it's only one item), and then running `Promise.all` on that array, something like this to prep:

```jsx
const attachmentURLs = dataAttachments ? Array.isArray(dataAttachments) ? dataAttachments : [dataAttachments] : [];
```

Then, you would just run your Array logic on `Promise.all(dataAttachments)`.

Another minor thing is to use a simple `forEach` instead of the `for` look inside of the then. The resulting code ends up a bit cleaner not having to juggle `i`:

```jsx
responses.forEach(response => {

});
```

rhauckdot:
Good call on converting to an array (I should have thought of that!). I added the requested documentation and converted the loops to forEach (old habit).

PR #1072, opened 2020-07-21
cms-warmup script improvements

frabarz:
* improves progress view
* fixes auth parameters
* adds new parameter: `--output, -o <filename>` : logs errors into that file in jsonlog format


frabarz:
Added a new command: `retry --input <path to file outputted by run> [--output <path for new error log>]`

PR #1073, opened 2020-07-23
[cms] Feature - Save to PDF

### LINKED ISSUE #1075: [cms] maintaining front-end profile page state with query args
There are 2 things I can think of:
- [x] selectors
- [x] tabs

jhmullen:
Adds save to PDF feature to profile pages.

- A new var called `print` is passed down through `context` from the top-level profile. It is `true` if `?print=true` in the query.
- The save button itself is in Hero.jsx, and is hidden in print mode in Hero.css.
- The "Appendix" vizes are actually "real" vizes with a `dataOnly` flag. This flag skips rendering the viz, and simply "passes through" an `Options.jsx` datapanel of the data. That means we could theoretically use this anywhere if someone wanted a pure data-table "Visualization"

There is a `Visible in PDF` dropdown in the SectionEditor.jsx, currently hooked up to nothing, as I've not changed the DB yet. Once the design pass is done (cc @davelandry) I will return [`i am jimdalf the white` 🧙‍♂️ ] and make the changes necessary for a migration.

Closes #1075


packages/cms/src/components/sections/Hero.jsx
```diff
@@ -75,6 +77,26 @@ class Hero extends Component {
...
+    if (!saving) {
+      const url = "/api/pdf";
+      const path = `${router.location.pathname}?print=true`;
+      const payload = {path};
+      const config = {responseType: "arraybuffer", headers: {Accept: "application/pdf"}};
+      this.setState({saving: true});
+      axios.post(url, payload, config).then(resp => {
+        const blob = new Blob([resp.data], {type: "application/pdf"});
+        const link = document.createElement("a");
+        link.href = window.URL.createObjectURL(blob);
```
davelandry, line 92: This "fake link" generation makes me nervous... just use `saveAs` like we do in `Options.jsx`!

https://github.com/Datawheel/canon/blob/master/packages/cms/src/components/Viz/Options.jsx#L142

packages/cms/src/components/sections/Hero.jsx
```diff
@@ -75,6 +77,26 @@ class Hero extends Component {
...
+        link.download = "your-file-name.pdf";
```
davelandry, line 93: Can you grab the Hero `title` and run it through this `filename` function from `Options.jsx`?

https://github.com/Datawheel/canon/blob/master/packages/cms/src/components/Viz/Options.jsx#L33-L35

PR #1076, opened 2020-07-28
[core] Lazy load image component

### LINKED ISSUE #1010: [core] create (or find) an Image component that uses web workers and lazy loads
My gut tells me that we'll want to roll our own React component, but if you can find an existing stable library that does these things and has an easy API, I'm open to that as well.

palamago:
New component exported by canon that resolve lazy load image in components that use background image.
Fixed #1010


palamago:
I agree! Good explanation of the process. Thanks.

About IE, I added all the polyfills that are -theoretically- needed. But couldn't test since my update to Catalina.

palamago:
- `background-image` or `img` supported.
- New config organization.

palamago:
- Added a `<noscript>` support for SEO reasons.

PR #1078, opened 2020-08-05
implements multi db on config

frabarz:
updates create-canon scaffold, fixes npx bin execution


examples/app/store/index.js
```diff
@@ -0,0 +1,23 @@
...
+  middleware.push(loggerMiddleware);
+}
+
+/**
+ * The object exported by this file should contain reducers to be
+ * combined with the internal default canon reducers.
+ */
+export const reducers = {
+  cms: cmsReducer
+};
```
davelandry, line 23: I really like reconfiguring all of the redux stuff into a `/store` directory. Can you write up migration steps for anyone updating their version of canon? (maybe just leave the write-up in this PR's main comment, and I'll copy/paste it into the release notes when ready)

examples/canon.js
```diff
@@ -19,9 +18,21 @@ module.exports = {
...
+      ]
+    },
+    {
+      host: env.CANON_DB_HOST || "localhost",
+      name: env.CANON_DB_NAME,
+      user: env.CANON_DB_USER,
+      pass: env.CANON_DB_PW,
+      tables: Object.values(cmsModelPaths)
+    }
+  ]
```
davelandry, line 37: This worried me at first, because I thought that we were going to have to require all user sites to create `models.js` type files re-listing the tables, but now I see what's going on and this really fantastic! 🍻 

Similar to my comment about the redux changes, can you add documentation to the core README file on how to set up database connections using `canon.js`, and then write some migration notes in this PR to be used in the release notes?

examples/canon.js
```diff
@@ -1,10 +1,9 @@
-/**
-  Simple test middleware function
-*/
-function exampleReduxMiddleware(canonMiddleware) {
-  return canonMiddleware;
-}
+const {modelPaths: coreModelPaths} = require("@datawheel/canon-core/models");
+const {modelPaths: cmsModelPaths} = require("@datawheel/canon-cms/models");
```
davelandry, line 2: Can you add these dependencies to the installation steps for the CMS (...and eventually the logins module)?

examples/app/store/index.js
frabarz: the migration is pretty simple:

- Create `store/` folder inside `app/` on the application, and create a `index.js` file inside
- Move the default exported object from `app/reducers/index.jsx` to this file, use `export const reducers =`
- Move the default exported object from `app/store.js` to this file, use `export const initialState =` (notice case)
- If used, move the middleware from `canon.js` to this file. In canon.js the middleware was enabled with a function `(applyMiddleware, canonMiddleware) => StoreEnhancer`, where the user had to add the middleware needed to a list of core middleware, and return `applyMiddleware(...array of middlewares)`. The new way only needs to export a `const middleware = []` as an array of the user defined middleware; the internals will do the combination with the core middleware and the call to `applyMiddleware`. The example also shows a way to add a middleware only while running on development mode.
- Update the references to needed imported objects
- Delete the old `app/reducers/index.jsx` and `app/store.js` files.

examples/canon.js
frabarz: To migrate a database, the DB environment variables the app was using must be setup in a new connection detail object in the `canon.js` file, following the instructions on the documentation.

PR #1120, opened 2020-10-14
[cms] adds warning to broken dropdowns

jhmullen:
Fixing this while passing through.

TextCards have advanced overrides for `allowed`, for custom string-based alloweds. There is no way here to catch for user error, as the user may have chosen to write `variableThatWontExistUntilSome[[Selector]]Happens` as their variable.

However, Generators and Sections do not currently support custom alloweds. They could, but in the interest of simplifying the interface, these top-level entities shouldn't need custom alloweds (famous last words?).

image
(Current allowed for sections - no custom)

This PR simply adds a text warning if the variable has been deleted out from underneath the allowed. Though this situation is possible/correct for advanced setups, I am simply trying to help new users avoid pitfalls with messages like this.

image

### DIFF
--- a/packages/cms/src/components/editors/VariableEditor.jsx
+++ b/packages/cms/src/components/editors/VariableEditor.jsx
@@ -349,6 +349,7 @@ class VariableEditor extends Component {
           {/* visibility */}
           {["generator", "materializer", "visualization"].includes(type) &&
             <VisibleSelector
+              type="generator"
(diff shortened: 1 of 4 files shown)


davelandry (approved):
Oh! I completely forgot that they currently (incorrectly) show "Allowed" when this happens. This is a great update.

PR #1122, opened 2020-10-15
[core] code splitting

davelandry:
This PR does the following:
* fixes express gzip compression by moving the middleware higher in the stack (995c043e)
* incorporates the preRenderMiddleware refactor by @frabarz from his `code-splitting` branch  (783491ba)
* adds sideEffects fields to the package.json of cms, logiclayer, and vizbuilder to enable tree-shaking (18dfd24c and 582e447b)
* updates example App.jsx and authentication docs in README to use `componentDidMount` instead of the deprecated `componentWillMount` (9ebd8e2c)
* adds `@loadable` dependencies to core for code-splitting (582e447b)
* updates webpack and some webpack related dependencies in core (582e447b)
* updates react and some react related dependencies in core (582e447b)
* creates and exports new `chunkify` helper function in core (d45cf62a)
* updates the webpack pipelines and client/server rendering to enable code-splitting and chunk detection (6f9c3094)
* chunkifies `Builder` and `Profile` CMS exports (a002c6b0)
* adds Code Splitting documentation to README (e8c908f5)
* uses `JSON.parse` for initial state objects (5c6d11afef85a46fff9d4ef877e2a995c363b59a)

After review, approval, and rebase of this branch, the following packages will need to receive minor releases to enable code-splitting and tree-shaking:
- core
- cms
- logiclayer
- vizbuilder


packages/core/src/helpers/chunkify.js
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * @param {Promise} statement The import() statement for the given component. Example: import("@datawheel/canon-cms")
+ * @param {String} [name = "default"] The component name to import, needed for named exports.
+ */
+export default (statement, name = "default") => {
+  const resolveComponent = components => components[name];
+  const lazy = loadable(() => statement, {fallback: Loading, resolveComponent});
+  lazy.resolveComponent = resolveComponent;
+  return lazy;
+};
```
frabarz, line 13: In the `chunkify` function, the first parameter is the promise returned by the `import()` call. This promise is then wrapped as the result of a function that is called later, but with a value corresponding to the promise's result, which was calculated at some point much earlier.
From a runtime perspective, it makes no difference to call this with `chunkify(() => import("..."))` and replacing `() => statement` to `statement`, but this would allow the promise to be executed until it's really needed.

packages/core/webpack/prod.js
```diff
@@ -36,32 +37,55 @@ module.exports = [
...
-    devtool: "cheap-module-source-map",
     context: path.join(__dirname, "../src"),
     entry: {
-      app: [
+      client: [
+        "normalize.css/normalize.css",
+        "@blueprintjs/core/lib/css/blueprint.css",
+        "@blueprintjs/icons/lib/css/blueprint-icons.css",
+        "core-js/modules/es6.promise",
+        "core-js/modules/es6.array.iterator",
```
frabarz, line 47: I'd like to experiment with this in the future. Last week I had to investigate further on babel configs, and found there are ways to bundle these and other polyfills only if we're really using them. It's easier than we think.

packages/core/src/server.jsx
```diff
@@ -121,40 +127,89 @@ export default function(defaultStore = appInitialState, headerConfig, reduxMiddl
...
+                }
+
+                if (process.env.CANON_BASE_URL) {
+                  scriptTags = scriptTags.replace(/\/assets\//g, "assets/");
+                  styleTags = styleTags.replace(/\/assets\//g, "assets/");
+                }
+
+                const componentHTML = renderToString(jsx);
+
+                const serialize = obj => `JSON.parse('${jsesc(JSON.stringify(obj))}')`;
```
frabarz, line 210: Wouldn't be better to use jsesc only instead of along with JSON.stringify? [it has this feature](https://www.npmjs.com/package/jsesc#json)

packages/core/src/server.jsx
```diff
@@ -175,18 +230,22 @@ export default function(defaultStore = appInitialState, headerConfig, reduxMiddl
     <script>
       window.__SSR__ = true;
       window.__APP_NAME__ = "${ req.i18n.options.defaultNS }";
-      window.__HELMET_DEFAULT__ = ${ serialize(headerConfig, {isJSON: true, space: 2}).replace(/\n/g, "\n      ") };
-      window.__INITIAL_STATE__ = ${ serialize(initialState, {isJSON: true, space: 2}).replace(/\n/g, "\n      ") };
+      window.__HELMET_DEFAULT__ = ${serialize(headerConfig, {json: true})};
+      window.__INITIAL_STATE__ = ${serialize(initialState, {json: true})};
```
frabarz, line 234: `serialize` only has one parameter

packages/cms/src/index.js
```diff
@@ -1,22 +1,36 @@
...
+  Section,
+  PDFButton,
+  Selector,
+  Stat,
+  Viz,
+  StatGroup
+};
+
+const Story = chunkify(import(/* webpackChunkName: "cms-story" */ "./components/Story.jsx"));
+const StoryLanding = chunkify(import(/* webpackChunkName: "cms-story" */ "./components/StoryLanding.jsx"));
```
frabarz, line 27: For the reasons I explained in the chunkify function, these `import()` calls load all the components on page load:
image
This is the index page for the example app, which doesn't import any of these, but loads them by importing anything from `@datawheel/canon-cms`.

frabarz (commented):
I left some comments on things I noticed while running the example app. I think the chunkify's first parameter is the main issue at hand, because it forces loading all the chunks on any page.

packages/core/src/server.jsx
davelandry: Good catch, this was a hold over from swapping `serialize` with `jsesc` in my testing.

packages/core/src/helpers/chunkify.js
davelandry: Fixed this. The `chunkify` function now accepts the anonymous function as its first argument. Even on just the homepage, this cut the lighthouse Time to Interactive and Total Blocking Time scores in half!

packages/core/webpack/prod.js
davelandry: This got in here because Babel told me to (in order to get the new babel import plugin working): https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import#working-with-webpack-and-babelpreset-env

I still only barely understand this core-js stuff. I had tried to upgrade to using v3 at one point, but hit a bunch of problems and slowly backed away. Would love some help here in the future, because I think this is often tied to getting `regenerator-runtime` working, which would allow for async React class methods.

packages/cms/src/index.js
davelandry: Fixed to use the correct `() => import()` argument syntax.

packages/core/src/server.jsx
davelandry: So, writing this:

```jsx
const serialize = obj => `JSON.parse('${jsesc(obj, {json: true})}')`;
```

Actually causes a syntax error:

Screen Shot 2020-10-27 at 4 25 29 PM

So this was the combination I was able to get working ¯\_(ツ)_/¯

frabarz (approved):
Everything working alright!

PR #1123, opened 2020-10-16
fixes missing key error in LazyImage

davelandry:
@palamago seems to have just been an issue using `<>` instead of the more explicit `<Fragment>`?

Test it for me to double-check... but ¯\\\_(ツ)\_/¯


palamago:
Ah, ok. That was easy. Good to know that <> works differently.

PR #1129, opened 2020-11-04
Issue 1095 redirects

### LINKED ISSUE #1095: [cms] make ids redirect to slugs for SEO

jhmullen:
Adds dynamic redirects to canon-core, and implements them automatically in the cms. This will require a minor release of both.

In Core, `needs` can now return a `canonRedirect` object, which contains values to swap in to the routes vars, such as `:slug` or `:id`. Docs are included in canon-core README.

In CMS, all profile ids now 301 permanent redirect to vanity slugs, making use of the above code. 

Closes #1095

### DIFF
--- a/packages/cms/src/components/Profile.jsx
+++ b/packages/cms/src/components/Profile.jsx
@@ -27,8 +27,10 @@ class Profile extends Component {
     if (profile.error) {
       const {error, errorCode} = profile;
       if (errorCode === 301) {
-        const link = linkify(router, profile.redirectData, locale);
-        return <Redirect to={link} />;
+        // console.log("got", profile);
+        return <div>no</div>;
+        // const link = linkify(router, profile.redirectData, locale);
+        // return <Redirect to={link} />;
       }
       const errorMessages = {
         404: "Page Not Found"
--- a/packages/cms/src/components/Profile.jsx
+++ b/packages/cms/src/components/Profile.jsx
@@ -5,6 +5,8 @@ import {fetchData} from "@datawheel/canon-core";
 import PropTypes from "prop-types";
 import ProfileRenderer from "./ProfileRenderer";
 import {NonIdealState} from "@blueprintjs/core";
+import {Redirect} from "react-router";
+import linkify from "../utils/linkify";
 import "./Profile.css";
 
 class Profile extends Component {
@@ -20,10 +22,14 @@ class Profile extends Component {
   }
 
   render() {
-    const {profile, formatters, locale} = this.props;
+    const {profile, formatters, locale, router} = this.props;
 
     if (profile.error) {
       const {error, errorCode} = profile;
+      if (errorCode === 301) {
+        const link = linkify(router, profile.redirectData, locale);
+        return <Redirect to={link} />;
+      }
       const errorMessages = {
         404: "Page Not Found"
       };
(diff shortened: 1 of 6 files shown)


packages/cms/src/components/Profile.jsx
```diff
@@ -5,6 +5,8 @@ import {fetchData} from "@datawheel/canon-core";
 import PropTypes from "prop-types";
 import ProfileRenderer from "./ProfileRenderer";
 import {NonIdealState} from "@blueprintjs/core";
+import {Redirect} from "react-router";
+import linkify from "../utils/linkify";
```
davelandry, line 9: why are these getting imported here? I don't see theme used in this file...

packages/core/src/server.jsx
```diff
@@ -147,6 +147,29 @@ export default function(defaultStore = appInitialState, headerConfig, reduxMiddl
             preRenderMiddleware(store, newProps)
               .then(() => {
 
+                // Needs may return a special canonRedirect key. If they do so, process a redirect, using the variables provided
+                // in those objects as variables to substitute in the routes.
+                const redirects = Object.values(store.getState().data).filter(d => d.canonRedirect);
```
davelandry, line 152: Instead of running `store.getState()` here, can you grab the `initialState` variable that also does that and move it above this code? If anything, we should probably be using that `initialState` code block intertwined here with your new code because it's also used down there to process any error codes in a need.

packages/core/src/server.jsx
```diff
@@ -147,6 +147,29 @@ export default function(defaultStore = appInitialState, headerConfig, reduxMiddl
...
+                // If the query contains ?redirect=true, a redirect has already occurred. To avoid redirect loops, ensure this value is unset
+                if (!req.query.redirect && redirects.length > 0) {
+                  // If any needs provided redirect keys, combine them into one object.
+                  const variables = redirects.reduce((acc, d) => ({...acc, ...d.canonRedirect}), {});
+                  // Use variables given by the canonRedirect key, but fall back on given params (to cover for unprovided keys, like :lang)
+                  const params = {...props.params, ...variables};
+                  // Not sure if this is a reliable way to get which route this is.
+                  let route = props.routes[1].path;
```
davelandry, line 160: I believe this is correct, because `props` is coming from the react-router `match` function.

packages/core/src/server.jsx
```diff
@@ -147,6 +147,29 @@ export default function(defaultStore = appInitialState, headerConfig, reduxMiddl
...
+                  Object.keys(params).sort(a => (/\d/).test(a) ? -1 : 1).forEach(key => {
+                    if (route.includes(`(/:${key})`)) {
+                      route = route.replace(`(/:${key})`, params[key] ? `/${params[key]}` : "");
+                    }
+                    else if (route.includes(`:${key}`)) {
+                      route = route.replace(`:${key}`, params[key]);
+                    }
+                  });
```
davelandry, line 168: I'm not sure why you're doing that number sort, but...

\* cracks knuckles *

```jsx
for (let key in params) {
  const pattern = `[(]{0,1}\/:${key}[)]{0,1}`;
  const result = data.replace(new RegExp(pattern, "g"), params[key]);
}
```

davelandry (changes requested):

davelandry (approved):

PR #1149, opened 2020-12-09
[cms] Feature local images

### LINKED ISSUE #1147: [cms] Local image hosting
Set the default behavior of images to host in the psql table as a blob. If cloud is enabled, default to there.

There will also need to be a resolution strategy for split-brain image hosting due to misconfiguration. @davelandry suggested that if an image is hosted locally but cloud is configured, upload it and delete the local when the image is requested. I am very hesitant to implement this, for two reasons. 

1) Performing a write operation as part of a read operation, especially when tied to a rate-limited API, is dangerous
2) A misconfigured cloud token could fail in an unexpected way, resulting in a deleted blob and un-uploaded cloud, with no clear path to recovery. 

For these reasons, I believe it's best to make the image fixing on-demand for cms admins. 

Progress: 
- [x] create splash/thumb columns
- [x] refactor upload path to fall back on psql upload
- [x] refactor imgserve path to respond with image buffer from psql
- [x] create reconciliation method, i.e., "this is local. send to cloud? y/n"
- [x] migration script for 0.17

Note: will require a full dot release and db migration to make room for img columns.

jhmullen:
Adds local image hosting in psql database, with reconciliation for split-brain installs.

closes #1147 

Requires a db migration, script included. will need a 0.17.0 release.


packages/cms/src/member/MetaEditor.jsx
```diff
@@ -745,7 +804,7 @@ class MetaEditor extends Component {
                 {currentRow.imageId && currentRow.dimension && currentRow.id && currentRow.cubeName
                   ? <img
                     className="cms-meta-selected-img"
-                    src={`/api/image?dimension=${currentRow.dimension}&cubeName=${currentRow.cubeName}&id=${currentRow.id}&type=thumb&t=${epoch}`}
+                    src={`/api/image?dimension=${currentRow.dimension}&cubeName=${currentRow.cubeName}&id=${currentRow.id}&size=thumb&t=${epoch}`}
```
davelandry, line 807: `type` changed to `size`?

1. I think you missed the `type=splash` in the Hero.jsx component
2. this change should be mentioned in the release notes

davelandry (changes requested):

packages/cms/src/member/MetaEditor.jsx
jhmullen: Nice catch - This is actually a bug, not a change. The imageRoute has always had two query args, `size` and `type`. `size` can be thumb/splash and `type` can only be `json`. 

In the CMS I was *incorrectly* using `type` for `size`, which simply meant that the preview images that were coming through were splash size (unnecessarily). The `type=splash` in the hero component is working by accident basically. 

No docs change needed, just need to invoke my own damn endpoint correctly. change incoming.

davelandry (approved):

davelandry:
@jhmullen changes approved 🎉 

I'll let you manage the merge/release, because of this being a major release and needing a version bump and release notes. Make sure you create a legacy branch of the current version before merging.

PR #1183, opened 2021-03-30
[cms] adds visible profile check to warmup script

### LINKED ISSUE #1182: [cms] catch for disabled profiles in cache script
CMS editors often use a (somewhat hidden) feature where they set `{visible: false}` on a search member's attribute content (cc @jhmullen). These profiles do not get shown in search results, and they should be filtered out of the cache warming script (as they will often produce N/As).

frabarz:
Closes #1182


frabarz:
changes applied to #1201, closing this PR

PR #1190, opened 2021-04-14
hotfixes table loading by removing bad control flow

jhmullen:
A faulty if/then prevented straight data from loading in table.jsx

### DIFF
--- a/packages/cms/src/components/Viz/Table.jsx
+++ b/packages/cms/src/components/Viz/Table.jsx
@@ -46,15 +46,13 @@ class Table extends Component {
     const url = config.data;
 
     if (url) {
-      this.setState({error: false, loading: url});
+      this.setState({error: false, loading: true});
       dataLoad.bind({})(url, dataFormat, undefined, (error, data) => {
-        if (JSON.stringify(this.state.loading) === JSON.stringify(url)) {
-          if (error) {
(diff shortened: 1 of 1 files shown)


packages/cms/src/components/Viz/Table.jsx
```diff
@@ -46,15 +46,13 @@ class Table extends Component {
     const url = config.data;
 
     if (url) {
-      this.setState({error: false, loading: url});
+      this.setState({error: false, loading: true});
       dataLoad.bind({})(url, dataFormat, undefined, (error, data) => {
-        if (JSON.stringify(this.state.loading) === JSON.stringify(url)) {
```
davelandry, line 51: @jhmullen removing this check actually re-introduces a potential race condition... and I'm not sure why this wouldn't work for data arrays. Maybe setting `loading: url` should be `loading: JSON.stringify(url)`? Here's the race condition this prevents:

1. component dispatches axios get for the `url`
2. before the data is loaded, the user clicks a selector that changes the value of `url`
3. component dispatches a 2nd axios get for the new `url`
4. the original `url` finishes loading, and sets state with that stale data
5. the new `url` finishes loading, and sets state again with the new data

So we're potentially now getting a 2nd React render and a flash of incorrect data.

PR #1207, opened 2021-05-18
[test] status API

palamago:
## New Package
I created a new package  called `canon-test` following the structure of the current packages and pnpm rules.
This new package is a super simple express app.

### To test it, run:
- `pnpm i` in root.
- Create a basic `.env` in `packages/test` -the same as you use in the examples page-. Remember to `source .env`.
- Terminal 1: `pnpm run dev` in `packages/test`.
- Terminal 2: `pnpm run dev` in `examples`.
- Go to: http://localhost:3000/status and see the results of the status test.

### CMS
- I created a new custom api `/api/alive/cms` that for now just return a `true` status. Could be more comprehensive in the future.

### Core
- I created a new custom api `/api/alive/core` that for now just return a `true` status. Could be more comprehensive in the future.

## Publish the package
Could you help me with publishing the new package?, I'm not sure how to use the `release.js`.
I need the new version numbers to test it in a real project. At least as a pre-release. I couldn't find a way to reference a dependency from a branch of a package that doesn't exists.

Thanks!


packages/test/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+    commands: await commands.run(),
+    apps:     await apps.run(),
+  };
+
+  //Send response
+  res.send(response).end();
+
+})
+
+const port = process.env.PORT || 3000;
```
davelandry, line 48: We probably want to namespace this variable... maybe `CANON_TEST_PORT`?

packages/cms/src/api/testCMS.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function(app) {
+
+  const {
+    CANON_API
+  } = process.env;
+
+  app.get("/api/alive/cms", (req, res) => {
```
davelandry, line 7: This comment is very nit-picky and applies to both`testCMS.js` and `testCore.js`: I think the word "alive" makes sense in this initial case, but if we are to eventually follow up with the `TODO` here I feel like this endpoint will be more of a "status" endpoint, not just checking if it's "alive". Maybe we make these endpoints `/api/status/cms` and rename the file to `status.js`?

packages/test/src/tests/ServicesChecker.js
```diff
@@ -0,0 +1,106 @@
...
+
+  constructor(ENV_VARS){
+
+    const {
+      CANON_CMS_CUBES,
+      CANON_DB_USER,
+      CANON_DB_PW,
+      CANON_DB_NAME,
+      CANON_DB_HOST,
+      CANON_DB_CONNECTION_STRING
```
davelandry, line 16: I'm not sure of an actual solution here, but: the names of these environment variables are actually not guaranteed. When we switched to having to hook up the models in `canon.js`, we removed the static association with these env vars (this is what enabled the OEC to hook into a 2nd db for the users table).

Since this is a canon plugin, we can probably assume that a `canon.js` file exists at the root folder. Maybe we can import that file somehow?

packages/test/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+const app = express();
+
+// Set the express middlewares
+app.use(express.static("public"));
+app.use(cors());
+app.use(compression());
+
+// Define the home route
+app.get("/", function (req, res) {
+  res.send("<h1>Canon Test!</h1>")
```
davelandry, line 25: Is your eslint plugin working in your IDE? I spotted this line missing an end `;`, but there may be a few other small things in these files.

davelandry (changes requested):

packages/cms/src/api/testCMS.js
palamago: Done! Good one.

packages/test/src/index.js
palamago: You are right. There was a problem with my eslint after some update. Solved.

packages/test/src/tests/ServicesChecker.js
palamago: Mmm, interesting. It is in the canon repo but is not exactly a canon plugin IMO. I'm still having an internal debate about it.
I can create some logic in the future that has a previous step first to validate the existence of certain required envvars and then run the actual validation. If some of the required envvars are not there I can return a status `[SKIPPED]` and the reason in string.

davelandry: @palamago can you make an issue in this repo for us to address this later?

punted! 🏈

davelandry:
@jhmullen this looks good to me. Can you give it a final lookover, merge, and release?

We're punting on one issue, which is that this script is checking for the existence of the old explicit DB env vars, before we put in the multi-db support and made hookups explicit in canon.js (check the unresolved issue thread).

packages/test/src/tests/ServicesChecker.js
palamago: https://github.com/Datawheel/canon/issues/1209

palamago:
BTW, we need a triple release:
`canon-core` 
`canon-cms`
`canon-test`

jhmullen (approved):

davelandry (approved):

PR #1215, opened 2021-06-04
[core] replaces react-hot-loader with react-refresh

frabarz:
This PR replaces canon-core's hot module reloading mechanism, from `react-hot-loader`, to React's Fast Refresh. The support now it's out of the box for the development environment.

It also fixes the packaging issues, where hot-loader was being packaged on production and in both targets, and removes all references to `react-hot-loader` from other modules in the monorepo.


packages/core/webpack/config/loaders.js
```diff
@@ -77,8 +82,12 @@ module.exports = props => {
     );
   }
   else {
+    if (!props.server) {
```
davelandry, line 85: @frabarz should this be `!props.server && !production`, or is it okay for this plugin to be part of the prod build?

davelandry (changes requested):
47 files changes 😭

40 of the files are cleanup and linting fixes 😍

This works fantastic @frabarz, it's so nice to strip out code and make this work automatically. Thank you for the changes made to `commonLoaders` as well.

When I ran `pnpm i` locally, it made some updates to `pnpm-lock.yaml` that looked necessary after I reviewed them. I pushed the changes to this branch, so take a look at that and see if those changes make sense (along with my 1 other comment on the babel plugin).

packages/core/webpack/config/loaders.js
frabarz: uh, look again at the `if` preceding this `else` block

frabarz:
I don't usually push `pnpm-lock.yaml` to feature commits because it's common they create a conflict at the moment of merging, and that results in an extra meaningless "Merge branch from..." commit. I think it'd be better to update that file before releases.
The `commonLoaders` function was something I was thinking for a while. It was using an envvar, so it wasn't a pure function, and that's something we should try to have into account in the future, so we have deterministic returned values and can write tests.

packages/core/webpack/config/loaders.js
davelandry: 🤦🏻 my bad...

davelandry (approved):

davelandry:
@jhmullen this all looks good to me, and it's ready for a merge and full major canon-core release 🚀

PR #1246, opened 2021-07-28
GDPR support

### LINKED ISSUE #685: [core] add GDPR opt-out
After some research, we should _probably_ (it's still slightly a grey area) add an "opt-out" style disclaimer directly in canon that displays if any of the following services have been enabled:
- [ ] google analytics
- [ ] google tag manager
- [ ] facebook pixel
- [ ] hotjar

I found the following library, which includes a ton of logic for this (but may be bloated for our needs): https://github.com/AmauriC/tarteaucitron.js

davelandry:
This PR closes #685 by adding 2 new environment variables:
* `CANON_GDPR` - setting this to `true` will show a minimally styled fixed position box on the bottom of the screen (shown here in screenshot) that will ask the user to accept/deny the use of tracking code (all text is stored in the locale files for translation/modification). Tracking code is only run when the user clicks "Accept", or they refresh the page (the lack of response is considered "consent").
* `CANON_GDPR_WAIT` - setting this to `true` requires a user to click "Accept" in order to run the tracking code, resulting in the pop-up being shown on every page visit until they either accept or reject.

Screen Shot 2021-07-28 at 4 16 32 PM

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -852,6 +852,8 @@ export CANON_LANGUAGE_DEFAULT="es"
 |`CANON_BASE_URL`|If hosting assets or running the server from a different location that the project folder, this variable can be used to define the base URL for all static assets. A `<base>` tag will be added to the start of the `<head>` tag.|`undefined`|
 |`CANON_GOOGLE_ANALYTICS`|The unique Google Analytics ID for the project (ex. `"UA-########-#"`). This also supports comma-separated values, if it's desired for pageviews to be reported to multiple analytics properties.|`undefined`|
 |`CANON_FACEBOOK_PIXEL`|The unique Facebook Pixel ID for the project (ex. `"################"`).|`undefined`|
+|`CANON_GDPR`|When set to `true`, tracking services like Google Analytics and Faceboook Pixel will only be loaded after the user accepts the usage of cookies and tracking. A pop-up Drawer will be shown on the bottom of the screen containing text and buttons that are editable in the locales JSON file.|`false`|
+|`CANON_GDPR_WAIT`|By default, if a user has not chosen whether to accept or reject the GDPR message, we will assume consent and not ask on subsequent page visits. If this env var is set to `true`, tracking services will only execute after the user has explicitly accepted.|`false`|
 |`CANON_GOOGLE_TAG_MANAGER`|The unique Google Tag Manager ID for the project (ex. `"GTM-#######"`).|`undefined`|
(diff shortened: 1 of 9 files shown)


frabarz:
I'm not completely clear on the difference on both env vars.
Also, the text is kinda confusing on the way the rule works: the user gives a tacit agree if they continues using the page after seeing the warning, but they are also given an accept/reject option. Is a mix of the european and american way to get cookie consent.

davelandry:
Here's the 3 different scenarios (there is only 3, because `CANON_GDPR_WAIT` has no effect without `CANON_GDPR = true`:

## Scenario 1 (Current/Default)
`CANON_GDPR = false`
`CANON_GDPR_WAIT = false`

### Behavior
1. No GDPR banner is shown.

## Scenario 2 (Passive Acceptance)
`CANON_GDPR = true`
`CANON_GDPR_WAIT = false`

### Behavior
1. On first visit to the site, the user is shown the GDPR banner and can perform 1 of 3 actions:
   - if they **Accept**, all of the currently installed services (Google Analaytics, Tag Manager, FB Pixel, and/or Hotjar) will execute, and continue to execute on every page load for 13 months until the cookie expires and they are asked again.
   - if they **Reject**, no installed services will run, and will not run on any subsequent page views for 13 months, until the cookie expires and they are asked again.
   - if they **Do Nothing**, and continue to browse the site without dismissing the pop-up, we do not _wait_ for them to give us permission, and the site will not show a GDPR banner on any subsequent full page load, and all services will execute for 13 months, until the cookie expires and they are asked again.

## Scenario 3 (Wait for Acceptance)
`CANON_GDPR = true`
`CANON_GDPR_WAIT = true`

### Behavior
1. On first visit to the site, the user is shown the GDPR banner and can perform 1 of 3 actions:
   - if they **Accept**, all of the currently installed services (Google Analaytics, Tag Manager, FB Pixel, and/or Hotjar) will execute, and continue to execute on every page load for 13 months until the cookie expires and they are asked again.
   - if they **Reject**, no installed services will run, and will not run on any subsequent page views for 13 months, until the cookie expires and they are asked again.
   - if they **Do Nothing**, and continue to browse the site without dismissing the pop-up, the site will continue to show a GDPR banner on any subsequent full page load, and no services will execute until they explicity "Accept" the conditions.

davelandry:
@frabarz and yeah, I also think the language of the text and buttons could be better, but I couldn't think of anything 🙁 

I pretty much just tweaked what the default text is in the library I'm using: https://github.com/Alex-D/Cookies-EU-banner

jhmullen (approved):

PR #1262, opened 2021-10-01
adds missing d3plus-text dependency

frabarz:


davelandry:
@jhmullen can I chuck this to you for a `canon-core` release?

PR #1279, opened 2021-10-29
adds xls and json downloads

jhmullen:
closes #1279


packages/cms/src/components/Viz/Options.jsx
```diff
@@ -77,35 +84,48 @@ class Options extends Component {
     }
   }
 
-  onCSV() {
+  onDownload(type) {
+    console.log(type);
```
davelandry, line 88: 😱

jhmullen: oh dang frickin flabbit

Issue #1284, opened 2021-11-09
[cms] Improve bilaterals legacy search


palamago:
- [x] New parameter to `ProfileSearch` that show more results from bilaterals
- [x] Change `/profilesearch` to include bilaterals combinations using the top 5 by zvalue

PR #1285, opened 2021-11-09
Improve search for bilateral profiles #1284

palamago:
Fixed: https://github.com/Datawheel/canon/issues/1284

UDD project is using the legacy search and they have a bilateral profile so we need to improve some results.

- New prop in `ProfileSearch` that forces to show more results from bilaterals.
- Change in `api/profilesearch` to include bilaterals combinations using the top 5 by `zvalue` 
- Readme update with the new prop

SORRY: some ESLint and Prettier formatting was applied in the files I changed.

### DIFF
--- a/packages/cms/README.md
+++ b/packages/cms/README.md
@@ -1,35 +1,37 @@
 # Canon CMS
+
 Content Management System for Canon sites.
 
 ## Table of Contents
-* [Why?](#why)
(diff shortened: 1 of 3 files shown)


davelandry:
👋🏻 just passing through here and wanted to give a 👍🏻 to the Prettier formatting. Could we even start standardizing this across projects? 🤔

PR #1326, opened 2021-12-20
adds custom resolve function to postcss

### LINKED ISSUE #1283: [core] enable aliases in CSS @import
Currently the aliases `$app` and `$root` are available when using `import` in javascript, but don't work in `@import` declarations in CSS files.

frabarz:
The [resolve parameter](https://github.com/postcss/postcss-import/tree/12.0.1#resolve) in `postcss-import` allows a custom resolve procedure, using node's native resolve function. This commit adds the possibility to resolve the `$app` and `$root` aliases, set in webpack.
Still, a more streamlined way, using the same configuration object, would be preferred.

Closes #1283

### DIFF
--- a/packages/core/webpack/config/common.js
+++ b/packages/core/webpack/config/common.js
@@ -1,0 +1,25 @@
+const path = require("path");
+
+const appDir = process.cwd();
+const appPath = path.join(appDir, "app");
+
+/** @type {import("webpack").ResolveOptions} */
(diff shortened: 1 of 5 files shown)


davelandry (approved):

davelandry:
@jhmullen this is good for any upcoming patch `core` release, when you have a moment 🚀

PR #1331, opened 2022-01-12
Simple UI Static Selector

greenrhyno:
Added Simple UI for creating a STATIC (aka non-dynamic) selector.

### DIFF
--- a/packages/reports/src/reports/blocks/Block.jsx
+++ b/packages/reports/src/reports/blocks/Block.jsx
@@ -54,6 +54,8 @@ function Block({id, modified, callbacks, inline}) {
   /* redux */
   const localeDefault = useSelector(state => state.cms.status.localeDefault);
   const blocks = useSelector(state => state.cms.reports.entities.blocks);
+
+  /** Block data object from database */
   const block = blocks[id];
   // todo1.0 fix formatters, maybe move this
   const formatters = useSelector(state => state.cms.formatters);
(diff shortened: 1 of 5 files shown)


packages/reports/src/reports/blocks/Block.jsx
```diff
@@ -211,7 +214,32 @@ function Block({id, modified, callbacks, inline}) {
...
   const executeButton = <Button style={{minHeight: 40}} onClick={() => onSave(true)}>Save &amp; Execute</Button>;
 
-  const components = {blockPreview, apiInput, textEditor, codeEditor, blockSettings, executeButton, simpleUI};
+  const isStatlike = ![BLOCK_TYPES.GENERATOR, BLOCK_TYPES.VIZ, BLOCK_TYPES.SELECTOR].includes(block.type);
+
+  /** This is the (non-code) GUI editor for blocks that is meant for most simple cases.
+   * If the block is a stat-like type, then use a rich text editor to render its forms.
+   * Otherwise, a block type will need to have its own tailored Simple UI logic.
+   */
+  const uiEditor = isStatlike
```
jhmullen, line 223: As we discussed in our call, the RichTextEditor should be merged into the SimpleUI umbrella, as it is just a special case of simple. But this is obviously fine for this PR

packages/reports/src/reports/editors/simple/SelectorUI.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const {onChange, locale} = props;
+
+  const [selectorIdentifier, setSelectorIdentifier] = useState("");
+  const [selectorType, setSelectorType] = useState(DEFAULT_SELECTOR_TYPE);
+  const [options, setOptions] = useState([]);
+  const stateFields = [selectorIdentifier, selectorType, options];
+
+  useEffect(() => {
+    setSelectorIdentifier(props.simpleState?.name || "");
```
jhmullen, line 30: This is a nitpick, but I prefer to destructure props in a single location, so that I don't have to chase around my inline usage of `props.var`. Could `simpleState` be pulled out alongside onChange and locale?

jhmullen (changes requested):

packages/reports/src/reports/blocks/Block.jsx
greenrhyno: Yep, that is part of my task for my next PR.

packages/reports/src/reports/editors/simple/SelectorUI.jsx
greenrhyno: You got it

greenrhyno: Change made.

PR #1337, opened 2022-01-14
Issue 1334 Refactor Block Editor Flow

### LINKED ISSUE #1334: [reports] Refactor block component design for easier editor management
Currently, `Block.jsx` creates a number of state-editing callbacks and various types of editors that use these callbacks, then passes these components as props to its child container component `BlockEditor.jsx`.  This clutters `Block.jsx` and makes it instantiate many things that are not needed for the current state of the Block Editor. It also confuses the control flow of data.

Task: Make `Block.jsx` only responsible for the maintaining of local `blockState` (a working copy of what is saved in the database) then make `BlockEditor` handle the creation of BlockType-specific editors and their methods needed for updating `blockState`. Then, have its children editors (whether code editors or "simple" UI editors) implement the BlockType-specific logic for updating  the `blockState`. Also add a consistent, general way for specific block editors to validate their changes.

To do:
- [ ] Move all `blockState`-editing methods to `BlockEditor`
- [ ] Make the getting of block variables (`useVariables`) consistent
- [ ] Make RichTextEditor a type of SimpleUI-type component rather than a separate thing
- [ ] Have `BlockElement` handle `isValidWorkingState` state
- [ ] Pass `setIsValidWorkingState()` down to `BlockEditor`
- [ ] Wrap state-editing methods created in `BlockEditor` so that they automatically update `isModified` and `isValidWorkingState` variables

greenrhyno:
Closes #1334 

Improves handling of branching BlockType-specific logic by making the delegation of responsibilities of various components related to block editing clearer.

Changes include:
- simplification and consolidation of block content editing methods
- expansion of `SimpleUI` to choose between custom UI editors or a generic `RichTextEditor`
- reduction of passed props by use of shared selector hooks (e.g. `useVariables`, `useBlocks`)
- reorganization of special BlockType-specific logic in `BlockPreview` into separate `BlockPreviewAdapter` files
- addition of logic to allow editor components to set a `isValid` state to avoid the saving on invalid intermediate states
- addition of JSDoc types for clearer function signatures


greenrhyno:
working on merge :( will finish monday

greenrhyno:
> working on merge :( will finish monday

Finished 💪

packages/reports/src/utils/consts/cms.js
```diff
@@ -81,7 +81,7 @@ const BLOCK_CONTENT_TYPES = {
   AUTHOR: "author",
   FOOTNOTE: "footnote",
   PARAGRAPH: "paragraph",
-  SELECTOR: "selector",
+  SELECTOR: "selector", // TODO - move to logic types
```
jhmullen, line 84: Let's either:
1) change this to `todo1.0` which has been my standard so far
or, if `TODO` is a thing that is supported by github/typescript/vscode/your own heart
2) change all `todo1.0`'s to `TODO` and stick with that from now on.

jhmullen (changes requested):

PR #1341, opened 2022-01-26
redesigns generator editor, API input/output, and variable lists (closes #1335)

### LINKED ISSUE #1335: [reports] Style inline generators
In cms-1.0-develop, one can click a generator in VariablesList, and the entire modal will replace its contents with the generator the user clicked. They can edit this, then click "Save & Return" at which point the modal returns to the statlike.

Dave, can you style this such that:
- [x] The hover is conditional over only generators
- [x] The hover is somehow meaningful for clicking (use a tooltip to explain?)
- [x] The transition of going from the normal editor to the `setInlineId` editors has some kind of visually helpful slide/transition
- [x] Any other design elements that would help the user understand "where we are"

davelandry:
@jhmullen I moved around a buuuuuunch of shit here, so please take a look and see if I'm doing everything correctly in this new `memoize` land please!

Screen Shot 2022-01-26 at 6 27 03 PM


jhmullen:
@davelandry I've got some crunchy wrapping columns when I test so you may need to have another look at the CSS:

image

packages/reports/src/reports/components/ApiInput.jsx
```diff
@@ -27,19 +34,59 @@ function ApiInput({defaultValue, id, onChange, onEnterPress}) {
...
+   * If no "blockStateContent" prop is given, defaults to the redux store value of block.
+   * Use the "blockStateContent" prop to provide override values if you want to show a
+   * live preview of unsaved changes.
+   */
+  const blockContent = blockStateContent || getBlockContent(block, locale);
+
+  const {duration} = useMemo(() => {
+
+    /** @type {import("./types/PreviewAdapters").BlockPreviewAdapterParams} */
+    const adapterParams = {active: true, block, blockContent, debug: true, locale, variables};
```
jhmullen, line 50: `duration` is only used by generators, as is `ApiInput`. The generatorPreviewAdapter should be used directly here (not using the lookup). It only has two arguments, `block` and `active`, so you shouldn't need all these arguments.

(separately, education-wise, your use of `locale` and `variables` in this adapterParams line would not necessarily be up to date, because they are not listed in the `[variables]` in the dependency argument)

Because you only need `block` and `active` here, you can remove the `blockStateContent` and `locale` that you added. Just directly use the generatorAdapter directly, the `block` from `useBlock`, and `active:true`

jhmullen (changes requested):

jhmullen:
A few more things:

- Please resolve the conflicts with ApiInput that are currently preventing a merge
- I like the inline API editor a lot - but I am missing the autocomplete feature for variables. If you have time, can the API bar use richTextEditor with some kind of auto-complete? I am glad to take this on if you just want to leave it as a vanilla textbar for now.

davelandry:
@jhmullen this is ready for a re-review! I think I'm starting to understand hooks... thank you for your patience 😅

davelandry:
@jhmullen I punted on ApiInput auto-complete for now... but wouldn't mind jumping in there and figuring it out if needed in a different PR

PR #1342, opened 2022-01-27
[core] Support Google Optimize

palamago:
- New envvar for Google Optimize code
- Add new script in the top of the page to load optimize script.


jhmullen (approved):

davelandry:
@jhmullen @palamago something important to note here, after looking at these code changes: the new `servicesHeadTags` to inject this new service into the head of the page circumvents the GDPR support I previously implemented, which is where `servicesScript` gets injected into the page via JavaScript if either GDPR is accepted or it is not enabled. If this Google Optimize service should respect GDPR, it really should be added into the pre-existing `servicesScript` variable.

PR #1343, opened 2022-01-28
Change google optimize load strategy to respect GDPR

palamago:
- Google optimize is loaded like the rest of the services.
- Respect GDPR
- JS code create a `<script>` tag at the very top of `<head>` (required)
- Fix bug in general click button handle that happens just with "accept gdpr" because ga exists but is not fully loaded yet.


davelandry:
@jhmullen this is approved, merged, and ready for the next canon-core release 🎉

PR #1346, opened 2022-02-01
Issue 1292 Dynamic Selectors

greenrhyno:
Adds the ability to create and edit dynamic selectors.

Dynamic selectors differ from "static" selectors in the sense that their list of possible options is derived from variables. This means that their list of options is "dynamic" and might change according to upstream Block events. Because of this, logic was added to account for the cases in where selectors' options changed and the selected value needed to be reset.

Changes include:
- added `status` actions to specifically add and remove query parameters from the URL
- added state management of `options` array in the `Selector.jsx` renderer
- added hook for editing invalid state in case of new options in `Selector.jsx`
- added edge case handling for dynamic selectors in `scaffoldDynamic.js`, `selectorQueryToVariable.js` and `runSelector.js`
- modified `SelectorUI.jsx` to allow for the creation of dynamic selectors
- added subcomponent for rendering a reactive form for specifying how to derive a Selector's options array from a dynamic variable


packages/reports/src/reports/blocks/types/renderers/Selector.jsx
```diff
@@ -18,17 +18,29 @@ export default function SelectorPreview({id, config}) {
...
+  const [serializedOptions, setSerializedOptions] = useState((config.options || []).map(d => d.id).join("-"));
 
-  const onChangeSingle = e => {
-    dispatch(setStatus({query: {[`selector${id}id`]: e}}));
-    setValue(e);
-  };
+  // change value if option list changes
+  useEffect(() => {
+    // options change ref when setStatus is called, so encode option content to compare equality 
+    const serializedNew = config.options.map(d => d.id).join("-");
```
jhmullen, line 27: Even though its tiny, we should extract out this "serializer" so it's not repeated in two places

packages/reports/src/reports/editors/simple/SelectorUI.jsx
```diff
@@ -107,22 +164,82 @@ function SelectorUI(props) {
...
+      optionsType === OPTION_TYPE.STATIC
+        ? <StaticSelectorOptionsEditor
+          staticOptions={options.static}
+          setStaticOptions={optionsState => setOptions({...options, static: optionsState})}
+        />
+        : <DynamicSelectorOptionsEditor
+          dynamicOptions={options.dynamic}
+          setDynamicOptions={optionsState => setOptions({...options, dynamic: optionsState})}
+          id={id}
+        />
```
jhmullen, line 183: I wonder if we could develop some sort of standard for conditionally rendering components like this. I have seen both:

- Declare the comps in a lookup object, before the render, and pull one out and insert it as `{myComponent}`.
- Declare the lookup object INLINE. Dave did this here once:
image

Not sure which one I like. Just food for thought.

packages/reports/src/reports/editors/simple/SelectorUI.jsx
```diff
@@ -107,22 +164,82 @@ function SelectorUI(props) {
...
+
+  /** Change the values of an option at a specific index */
+  const editOption = useCallback(
+    (idx, newValues) => {
+      if (!newValues || idx >= staticOptions.length || idx < 0) return;
+
+      const newOptions = [...staticOptions];
+      newOptions.splice(idx, 1, {...staticOptions[idx], ...newValues});
+      setStaticOptions(newOptions);
+      // TODO - set error
```
jhmullen, line 206: We have some great toasts for errors now - check out `import {useNotifications} from "@mantine/notifications";`

jhmullen (approved):

PR #1354, opened 2022-02-25
Openid connect

palamago:
Effort to support OpenID connections as login strategy: https://openid.net/

- Implemented http://www.passportjs.org/packages/passport-openidconnect/ package
- Small refactors in the generic strategy methods.
- Add user login/logout buttons in the example app.
- New openId button, logo and translations for login form component.
- No database changes.


davelandry:
I'd like @jhmullen to review this one, as there needs to be more people in the company that understand how the login systems work (they're actually pretty simple), and he is the current maintainer of this repo.

Issue #1348, opened 2022-02-11
[reports] Refactor Block Editor Flow 


greenrhyno:
Currently, the components that are responsible for editing blocks are (quasi) controlled -- they are passed a `blockContent` object which represents the latest edited state of a block's content (like its js logic and other metadata). These block-editing components are also passed a setter (`setBlockContent`) which is used to change the object they are passed. This is causing a number of unnecessary renders to be triggered because this `blockContent` object is constantly changing (sometimes on keystroke!).

Refactor the data flow of these block editors and the `Block` data flow to avoid unnecessary rerenders and make the components of Blocks (renderer, simple ui, adapter) less reliant on lots of props from ancestors so that they are easier to implement as plugins in the future.

Some things to do:
- [ ] Add debouncer to `blockContent` so `setBlockContent` callback does not trigger so many rerenders
- [ ] Make simple ui editors uncontrolled so they don't rely on `blockContent` prop
- [ ] Don't unmount the "simple" editor when switching modes so they can maintain state when switching modes
- [ ] make BlockPreview take blockContent as a prop so it does not choose internally between redux state and local state
- [ ] make config objects for blocks that can configure behavior (like rendering preview on local state vs only on save)
- [ ] add execute button to simple UI if block type renders only on save

greenrhyno:
A diagram from a discussion about the new flow:
![IMG_0714](https://user-images.githubusercontent.com/38231289/153662762-676baf5c-5319-4bd9-aef4-cc0511ab706f.jpg)

PR #1358, opened 2022-03-11
1348 (part 1) Add Block Configs

greenrhyno:
Adds a BlockConfig object that helps simplify some of the differing logic and behaviors of different types of blocks. Currently, there exist many pieces of logic like "if the block type is Viz, do this" -- these changes are the beginning of a push to remove all those type-specific branches of logic and replace them with checks for a block's _behaviors_ rather than _types_. This generalization and abstraction should allow for more maintainable code and easier maintenance and extensibility.

Addresses some of the issues mentioned in #1348


greenrhyno:
@jhmullen I chose to leave some of the rendering refactors of the block editing for another day since I thought it was better to get these changes out there first.

packages/reports/src/reports/blocks/BlockPreview.jsx
```diff
@@ -36,40 +38,56 @@ function BlockPreview(props) {
...
+  const blockContent =  BLOCK_CONFIG[block.type].renderPreviewOnEdit && blockStateContent || getBlockContent(block, locale);
 
   /* redux */
   const formatterFunctions = useFormatters(locale);
 
-  const {content, error, log} = useMemo(() => {
-    let payload = {};
-    // if a Block-specific preview adapter function exists, use that to build payload
-    if (PreviewAdapters[block.type] && typeof PreviewAdapters[block.type] === "function") {
+  const [content, error, log] = useMemo(() => {
```
jhmullen, line 46: For my own edification, is there a performance or common practice reason for choosing an array instead of an object here?

packages/reports/src/reports/blocks/types/adapters/Selector.js
```diff
@@ -0,0 +1,37 @@
...
+  return options.map(d => typeof d === "string" ? {id: d, label: d} : d);
+};
+
+const maybeFixForMulti = (defaultValue, type) => 
+  type === SELECTOR_TYPES.MULTI && !Array.isArray(defaultValue)
+    ? [defaultValue]
+    : defaultValue;
+
+/** @type {import(".").BlockPreviewAdapterFunction} */
+export default (vars, id) => {
```
jhmullen, line 15: I am realizing that the longer I spend away from these functions, the more I forget what they do. In this case I realize the Selector function is an adapter - where is the proper place to document exactly what's happening here? I'd like to see a function description here - and perhaps a similar one in the other adapters. I'm a big fan of putting examples in the comment doc: "this function takes variables calculated by X thing and exports Y config, and here's an example of what that might look like"

I know for example this note to myself saved me a bunch of times: https://github.com/Datawheel/canon/blob/master/packages/cms/src/utils/varSwapRecursive.js#L42-L46

Wondering your thoughts on how to standardize these kinds of function recipe/descriptions?

packages/reports/src/reports/blocks/types/index.js
```diff
@@ -0,0 +1,85 @@
+import {BLOCK_TYPES} from "../../../utils/consts/cms";
+
+import adapters from "./adapters";
+import editors from "./simpleEditors";
+import renderers from "./renderers";
```
jhmullen, line 5: Just want to say that this little three-spot of imports makes me happy. I'm glad we talked this out on the whiteboard

jhmullen (approved):
Everything here looks good, I especially appreciate the comments about future dev. We need to move forward in small bursts, and writing notes to ourselves about forthcoming architecture is important.

Just a few conversational threads here! I'll be merging first thing Monday.

packages/reports/src/reports/blocks/BlockPreview.jsx
greenrhyno: Not a performance reason really -- it just allows you to destructure the variables under whatever name you want them rather than having to remap named variables in an object.

packages/reports/src/reports/blocks/types/adapters/Selector.js
greenrhyno: There is one! See the index.js file in this same directory. That `BlockPreviewAdapterFunction` typedef has a short description of what this function does, though it could be improved.

Basically, this is a function that is called to transform the output of a block's evaluated logic to the props of the BlockType's renderer React component. This is an optional thing where one might want to write logic for programming fallbacks or turning the output variables into something that is only needed for rendering.

PR #1377, opened 2023-02-20
First canon-next iteration

scespinoza:
Notes:
- added `Readme.md`


example-next/pages/profile/[...members].js
```diff
@@ -0,0 +1,64 @@
...
+export default ProfilePage;
+
+
+export async function getStaticPaths() {
+  return {
+    paths: [
+      // Place here the paths you want to cache on build
+    ],
+    fallback: true
+  };
```
Comment (davelandry) [inline example-next/pages/profile/[...members].js:47]: Instead of requiring sites to implement the entirity of `getStaticPaths` (but I understand the need for this export to still exist in the file), can we export a helper function that accepts an optional paths argument? That way if we need to change some of these defaults, we won't necessarily require all sites to copy/paste the new settings:

```
return profileStaticPathsDefaults(); // I'm open to suggestions about a better/shorter function name
```

example-next/pages/profile/[...members].js
```diff
@@ -0,0 +1,64 @@
...
+  const formatters = await axios.get(formatterURL.href)
+    .then(resp => resp.data);
+
+  return {
+    props: {
+      ...await serverSideTranslations(locale, ["common", "profile"]),
+      formatters,
+      profile
+    }
+  };
```
Comment (davelandry) [inline example-next/pages/profile/[...members].js:63]: Similar comment from `getStaticPaths`, can we move `getProfile` and all of this logic into:

```
return profileStaticPropsDefaults(); // I'm open to suggestions about a better/shorter function name
```

packages/next/README.md
```diff
@@ -1 +1,71 @@
...
+
+This package provides React Components for rendering a canon profile on a NextJS app, using Mantine components.
+
+# Setup an installation
+
+You'll need to create a NextJS (v13) app. Install the package using npm:
+
+```npm i @datawheel/canon-next```
+
+You'll need to wrap your entire app into a `MantineProvider`, in order to allow Mantine components to render properly. For detailed instructions follow the official [Mantine Documentation for NextJS](https://mantine.dev/guides/next/). In adition, you'll need to set up `next-i18next` on your app, with the appropiate locales.
```
davelandry, line 11: spelling: "adition" to "addition"

Also, could you put a very simple example here of what `_app.js` would look like on initial install, with the providers and translations set up?

packages/next/README.md
```diff
@@ -1 +1,71 @@
...
+
+If you chose to serve the profile pages from a path other than `/profile`, you'll need to provide an appropiate `linkify` function. The `linkify` function should take an array of slug/id pairs and convert it to a valid url path in your app.
+
+
+## Aplying custom styles
+We recommend using Mantine components overrides into the `theme` object of the `MantineProvider`, as described in the [mantine documentation](https://mantine.dev/theming/theme-object/). If you need more fine grained adjustements, you can target some components using its class name through the `globalStyles` object of your `MantineProvider`
+
+
+
+```
```
davelandry, line 40: when writing code blocks, you can get better syntax highlighting by specifying the language immediately after the initial three back-ticks, try throwing a `jsx` in here on the same line as the first set of backticks.

https://github.github.com/gfm/#example-112

davelandry (changes requested):

packages/next/README.md
scespinoza: added also a link to our template, that has all already installed

PR #1380, opened 2023-05-18
Canon next 0.2.0

scespinoza:
CHANGELOG:
- Implements turborepo and better tsup config for `example-next` and `packages/next` workspaces.
- Sets `React` explicit import on all files to avoid bundling errors and simplify transpiling.
- Enable profile comparisons through env var.
- Hook up selectors state to query string.
- Refactor of `ProfileRenderer` to increase performance of the react rendering cycle. A lot of the rendering logic now has its own separate custom hook instead of being all straight into the component.
- Improve filtering and loading states on `ProfileSearch`. We also export a `ProfileSearchModal` wrapped into a mantine modal, and a "vainilla" `ProfileSearch`.
- Condition `SegmentedControl `rendering on the length of the labels, as the component often breaks with long names.
- Refactor some buttons and headings to have a more modern look, using straight mantine components instead of the `Parse` function. Example:
Captura de pantalla 2023-05-18 a la(s) 16 25 13

- The profile renderer now accepts a `icons` mapping object between the legacy blueprint ids and any `ReactNode`.
Captura de pantalla 2023-05-18 a la(s) 16 29 05

- Better organization of DOM elements in `Hero`.

Notes:
- Still figuring out PDF prints, but I decided not to hold on with this as this isn't necessary on Monshaat nor OEC AFAIK. I'll probably bring it from bespoke eventually.
- Now you can run the developement environment from the root of the repo with just `npm run dev` my understanding is that this doesn't break the other workspaces as they use `pnpm` (but if it does, please let me know).
- There is a env var commented in the code that I couldn't find where it came from or what it is supposed to do:
`// const comparisonExclude = process.env.NEXT_PUBLIC_PROFILE_COMPARISON_EXCLUDE;`. If it is important please let me know.
- I also did some subtle changes on the nesting of some elements, taking care of semantics and SEO.
- I think this deserves to release as `0.2.0`


packages/next/cms/components/ProfileRenderer.js
```diff
@@ -4,462 +4,68 @@
...
 
-// TODO
-// User must define custom sections in app/cms/sections, and export them from an index.js in that folder.
-// import * as CustomSections from "CustomSections";
+import ComparisonButton from "./fields/ComparisonButton";
 
-// TODO
-// const comparisonsEnabled = process.env.NEXT_PUBLIC_PROFILE_COMPARISON;
+const comparisonsEnabled = process.env.NEXT_PUBLIC_PROFILE_COMPARISON;
 // const comparisonExclude = process.env.NEXT_PUBLIC_PROFILE_COMPARISON_EXCLUDE;
```
davelandry, line 23: I know of the exact use case for this env var. I didn't approve it's inclusion in canon-cms, and I would have probably done it as a React prop or a CSS hack in the project, but InsightsCloud has this env var:

```sh
export CANON_CONST_PROFILE_COMPARISON_EXCLUDE="skills_and_occupation"
```
What this is doing is saying "don't show the comparison button for the following comma-separated profile slugs". As an example, `"country,hs"` could also be a valid value for a site like the OEC.

Might as well just include, shouldn't be too hard, right?

packages/next/package.json
```diff
@@ -54,7 +57,7 @@
     "next-i18next": "^13.1.5",
     "react": "^18.0.2",
     "react-inlinesvg": "^3.0.1",
-    "xlsx": "^0.18.5"
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.19.3/xlsx-0.19.3.tgz"
```
davelandry, line 60: Just curious, what's the deal with this?

packages/next/cms/components/Viz/Table.js
```diff
@@ -1,4 +1,6 @@
+import React from "react";
 
+/** TODO */
```
davelandry, line 3: I'm okay not doing this one for this release, but let's make sure we remember that we need it for OEC 🧠

packages/next/cms/hooks/use-comparison.js
```diff
@@ -0,0 +1,38 @@
...
+  const [comparisonLoading, setComparisonLoading] = useState(false);
+  const {query, locale} = useRouter();
+  const compareSlug = query?.compare;
+
+  const slug = profile.meta[0].slug;
+
+  useEffect(() => {
+    if (compareSlug) {
+      setComparisonLoading(true);
+      const url = `${process.env.NEXT_PUBLIC_CMS}profile?slug=${slug}&id=${compareSlug}&locale=${locale}`;
```
davelandry, line 20: Should this env var be declared as a `const` at the top of this file? (we probably should just test a prod build)

packages/next/cms/utils/formatProfileResponse.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const newImages = [];
+  if (dims) {
+    // eslint-disable-next-line no-plusplus
+    for (let i = 0; i < dims.length; i++) {
+      if (images[i]) {
+        newImages.push({
+          src: new URL(
+            `/api/image?slug=${dims[i].slug}&memberSlug=${dims[i].memberSlug}&size=splash`,
+            process.env.NEXT_PUBLIC_CMS
```
davelandry, line 22: Same as my other comment, should this be a `const` up above?

davelandry (changes requested):
In addition to my inline comments:
 - I'm perfectly fine with you removing any more `<Parse />` instances as you want, working towards getting rid of that component completely. It was made at a time when Blueprint couldn't do everything we wanted by a designer learning React.
 - ...can you help me run this locally? 😅 How should I be installing dependencies now?

packages/next/package.json
scespinoza: xlsx isn't being published on npm because they have some legal issues (?), we had to do the same and bespoke and I couldn't find an easy replacement for the library, but I'd love to replace it in the future

packages/next/cms/components/Viz/Table.js
scespinoza: didn't know we needed it for the OEC, noted

packages/next/cms/hooks/use-comparison.js
scespinoza: if I do 
```js
const NEXT_PUBLIC_CMS = process.env.NEXT_PUBLIC_CMS;
```
it gets transpiled by tsup to:
```js
const {NEXT_PUBLIC_CMS} = process.env;
```
 and next complains. There is probably some rule that I could override but couldn't find it. Also, monshaat already works like this with no issues.

davelandry:
@scespinoza I'm able to get it up and running (thank you), but I'm seeing some issues that I think may be related to the `NEXT_PUBLIC_CMS` variable? Should this be a valid `.env` file?

```sh
NEXT_PUBLIC_CMS=https://app-owl.oec.world/api
CANON_LANGUAGE_DEFAULT=en
CANON_LANGUAGES=en
```

When using this I get some React errors using the profile search on the base route. I also see in `ProfileSearch` and `ProfileTile` that we are using string concatenation with the env var, which assumes the env var does not have a trailing slash (so we should switch those to the `new Url` way, which catches for the double slash).

Screenshot 2023-05-24 at 11 56 02 AM

packages/next/cms/components/ProfileRenderer.js
scespinoza: Test with
```
NEXT_PUBLIC_TESSERACT=https://monshaat.api.dev.datawheel.us/tesseract/
NEXT_PUBLIC_CMS=https://monshaat.cms.dev.datawheel.us/
NEXT_PUBLIC_PROFILE_COMPARISON=true
NEXT_PUBLIC_PROFILE_COMPARISON_EXCLUDE="manufacturing"
```

davelandry: @scespinoza not quite the logic here, the exclusion is the _entire profile type_, not for specific members within. It's not something we would even really use right now in Monsha'at, because currently there is only 1 profile type in that project ("sector").

As an example, if a site had two different profiles, one for Countries (`/profile/country/<memberSlug>`) and another for Products (`/profile/hs/<memberSlug>`), and we wanted to enable comparisons for all current and future profile types _except_ for product profiles, we would set:

```sh
NEXT_PUBLIC_PROFILE_COMPARISON_EXCLUDE="hs"
```

It's the slug of the primary profile _dimension_, not an individual _member_ of that dimension.

davelandry:
@scespinoza using your latest env vars, profile search works but now it looks like the URL used when fetching the comparison is missing the `/api` part:

Screenshot 2023-05-26 at 9 56 25 AM

Also, can we change the `alert()` popup usage into an error that appears on the page itself? Maybe replace the "Add Comparison" button with an "Error. Please Retry" button that allows you to try and select a comparison again?

scespinoza:
@davelandry I fixed the comparisons call, and I noticed the api endpoint always returns a 200 response, even if it doesn't find the profile, just returns a error message in the body 🤔 It feels weird, but I handled it that way

packages/next/cms/components/ProfileRenderer.js
scespinoza: fixed

packages/next/cms/hooks/use-profile-sections.js
```diff
@@ -60,7 +60,7 @@ export default function useProfileSections(profile, comparison) {
   const comparisonSections = [];
 
   if (comparison) {
-
+    console.log("comparison", comparison);
```
davelandry, line 63: 💣

davelandry (changes requested):

davelandry:
@scespinoza in addition to my 💣 above, we also need to tweak the layout of the sections. Take a look at how it currently works in DataMexico and [Data USA](https://datausa.io/profile/geo/california?compare=florida):

Screenshot 2023-05-26 at 3 28 13 PM

It's much easier to compare when you `flex-direction: column;` the whole section and place it side-by-side with the comparison section. We also use the word `t("in")` when we modify the titles, which I think is nicer than the hyphen.

scespinoza:
@davelandry fixed the column layout. Didn't really get that coment about `t("in")`

davelandry:
@scespinoza let's goooooooooo

I will leave it for you to cut the `0.2.0` release 😄

PR #1389, opened 2024-04-22
Wb/1309 fe and css issues

walbuc:
Updates to profile search required for oec:

https://github.com/Datawheel/oec-site/issues/1309.

- Added controlled prop implementation for *opened* modal prop.
- Added auto focus custom hook for input search.

### DIFF
--- a/packages/next/cms/components/ProfileRenderer.js
+++ b/packages/next/cms/components/ProfileRenderer.js
@@ -111,7 +111,7 @@ function ProfileRenderer({
 
   return (
     <ProfileContext.Provider value={context}>
-      <div id="walter" className={`cp${print ? " cp-print" : ""}`}>
+      <div className={`cp${print ? " cp-print" : ""}`}>
 
         { !hideHero
           ? (
--- a/packages/next/cms/components/ProfileRenderer.js
+++ b/packages/next/cms/components/ProfileRenderer.js
@@ -44,6 +44,8 @@ function ProfileRenderer({
   customSections = {},
   relatedTile,
   t,
+  opened,
+  handlers,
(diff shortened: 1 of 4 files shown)


packages/next/cms/components/fields/ProfileSearch.js
```diff
@@ -277,6 +277,9 @@ export function ProfileSearch({
 
   const [debouncedQuery] = useDebouncedValue(query, 400);
   const profiles = useCMSProfiles();
+  const [inputRef, setInputFocus] = useFocus();
+
+  React.useLayoutEffect(() => setInputFocus(), [setInputFocus]);
```
scespinoza, line 282: Shouldn't this go inside the `useFocus` hook? (And we could steal this for bespoke 🥷🏻)

Also, this will run on every re-render, as `setInputFocus` gets created on every function call. Do we want that?

packages/next/cms/components/ProfileRenderer.js
```diff
@@ -44,6 +44,8 @@ function ProfileRenderer({
   customSections = {},
   relatedTile,
   t,
+  opened,
```
scespinoza, line 47: `searchOpened`/`searchHandlers`?

scespinoza (changes requested):

packages/next/cms/components/ProfileRenderer.js
walbuc: Updated props naming. to searchOpened & searchHandlers!

scespinoza (approved):

davelandry (approved):

PR #1390, opened 2024-04-26
simplify & focus with current FocusTrap mantine component

walbuc:
After testing with the latest release I found a weird focus behavior. The usefocus hook conflicted with the FocusTrap mantine component. So I opted in to use the current mantine implementation and found a simpler and elegant solution, that solves the input focus issue.


davelandry (approved):
Super simple, love it 👍🏻

